PR #9, opened 2022-08-10
Add CI

Zoxc:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+
+    - name: Build native-activity
+    working-directory: android-activity
+    run: >
+      cargo ndk
+      -t arm64-v8a
+      -t armeabi-v7a
+      -t x86_64
+      -t x86
+      build --features native-activity
```
MarijnS95, line 63: This doesn't seem indented properly.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+env:
+  CARGO_INCREMENTAL: 0
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
MarijnS95, line 17: Yaml style requires arrays to be indented as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: >
+        cargo ndk
+        -t arm64-v8a
+        -t armeabi-v7a
+        -t x86_64
+        -t x86
+        -o app/src/main/jniLibs/ -- build
+
+    - name: Format
+      run: cargo fmt --all -- --check
```
MarijnS95, line 76: Fwiw I'd perform cheap checks up-front, instead of build-testing the entire thing and then eventually rejecting it because the formatting is wrong.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+      working-directory: android-activity
```
MarijnS95, line 77: This is part of the workspace, just run this (and other commands) from the root. Same comment everywhere else this working directory is specified, if needed use `-p`.

.github/workflows/ci.yml
Zoxc: I prefer formatting errors to not hide more interesting errors. I split the formatting to a separate job so the formatting feedback is faster.

.github/workflows/ci.yml
Zoxc: I did specific the directory for clarity vs. the example workspaces, but it is indeed not required.

rib:
Ooh, cool thanks for taking a look at setting up CI!

.github/workflows/ci.yml
MarijnS95: I find `-p` clearer, but alas.

.github/workflows/ci.yml
MarijnS95: I prefer formatting errors to prevent more expensive jobs from starting. I've never seen a contributor fix both formatting errors _and other build/test/clippy failures_ in one go (even if the hope/dream is these standard checks are executed locally before even considering to open a PR).

rib:
I need to get up to speed with how github actions work (I've mainly used Travis or Gitlab runners for CI and haven't had a chance to play with actions so far) so I don't have any strong opinions here.

I didn't see an obvious way for test this out before landing but I figured we can iterate on this anyway so I've just gone ahead and merged.

Thanks for looking at this.

PR #11, opened 2022-08-11
Add minimal Cpal audio library example

rib:
Based on the android example that's in the cpal repo, this test plays
a 440hz sine wave and is based on GameActivity

_Note: this requires a workaround branch for cpal that comments out
a dev-dependency on ndk-glue that cpal has for its android example.
This is needed because Cargo is spuriously propagating this dev
dependency outside of the cpal package (which leads to a crash)_


rib:
While testing this I was hitting a crash when pausing/resuming the app which I was initially assuming was a oboe-related issue in cpal but it turned out to be an NDK 0.7 related issue that was actually affecting all the examples: 93828a18a9be395d7bd9bb660f4926f38b5fe85a

PR #15, opened 2022-08-12
Link the C++ standard library correctly.

Zoxc:
`libc++.a` is just a simple link script `INPUT(-lc++_static -lc++abi)`. I made the build script match so we link to the complete C++ standard library.

Fixes https://github.com/rib/android-activity/issues/10.

### DIFF
--- a/android-activity/build.rs
+++ b/android-activity/build.rs
@@ -29,6 +29,10 @@ fn build_glue_for_game_activity() {
         .extra_warnings(false)
         .cpp_link_stdlib("c++_static")
         .compile("libnative_app_glue.a");
+
+    // We need to link to both c++_static and c++abi for the static C++ library.
+    // Ideally we'd link directly to libc++.a.
+    println!("cargo:rustc-link-lib=c++abi");
 }
 
 fn main() {


rib:
I'm not sure we can make this assumption at this level - this is the kind of trick that probably belongs in build tools like `cargo ndk` and then it's somewhat dependent on the application which choice of libstd++ library is used, either static or shared.

E.g. `cargo apk` tries to automatically add the shared libc++_shared.so to your apk if you're dependent on it, and it would maybe be good if `cargo ndk` had some functionality to help with that too.

In general, this feels like a build system concern.

rib:
E.g. ref the various different c++ runtime options, that may all be valid choices, and should all be compatible with `android-activity`: http://android-doc.github.io/ndk/guides/cpp-support.html

Zoxc:
The build script already picks the static variant of libc++, so this PR is a strict improvement, because it does so correctly.

How to link against the C++ standard library should probably be specified with some environmental variable, but I don't think a standard exists for that.

Note that Rust crates which use C++ must themselves link against the C++ standard library in their build script.

Both `cargo apk` and `cargo ndk` should not themselves add a dependency on the C++ standard library because they're made to build Rust, not C++.

rib:
okey, had a bit more time to think about this...

yeah, I suppose the fact that there's currently no standard established in the rust ecosystem for how to coordinate which c++ runtime libs to use then we have to handle it more ad-hoc for now.

re `cargo apk/ndk` I was thinking in the sense that they could potentially coordinate a consistent choice across crates (they do effectively compile c/c++ code too) comparable to `ANDROID_STL` in cmake, or an environment variable as you mentioned. Things break when you have different components linking with different standard libraries so the current situation isn't exactly ideal.

As you say though, since this crate does itself use c++ so we do of course need to say something about linking to the standard library (currently `c++_static`) - though normally I'd only expect to be choosing `c++_shared` vs `c++_static` on android for the standard library.

For reference; the `cc` crate would normally link with the `c++_shared` standard library automatically by default (with a `CXXSTDLIB` environment variable to override) and I mainly just specified `c++_static` for convenience to avoid needing to manually copy `libc++_shared.so` across when building an apk.

It's also notable that the `cc` crate specifically lists `c++_shared` and `c++_static` as valid options for Android, but since you found that  `libc++.a` is a linker script for both libc++_shared.a and libc++abi.a it could make sense for us to also file an issue with the `cc` crate if it needs to be more aware of how to correctly link the static standard library on Android.

Looking at some of the Android NDK docs I noticed they highlight here: https://developer.android.com/ndk/guides/cpp-support#use_clang_directly that if using clang there is a specific `-static-libstdc++` option for selecting the static version of the library.

I guess atm that some of the Android docs about `c++_shared` vs `c++_static` have lead to a common misconception that those are the library names you're suppose to actually use when linking.

Incidentally looking at `android-ndk-r25/build/cmake/android-legacy.toolchain.cmake` they have:
```
elseif(ANDROID_STL STREQUAL c++_static)
  list(APPEND ANDROID_LINKER_FLAGS "-static-libstdc++")
```

I have a feeling that the `cc` crate should be doing something similar to link with `libc++.a` Actually I see they do even have a `static_crt()` api intended for msvc: https://docs.rs/cc/1.0.73/cc/struct.Build.html#method.static_crt which would potentially make sense to support with clang too.

I think for now, this fix is reasonable from a pragmatic pov; though I do tend to think it feels like a workaround, which will make things more fiddly later for anyone wanting to instead link with the shared library.

Maybe the better solution to aim for longer term will be to file an issue with the `cc` crate and see if that can be updated to know how to correctly link with the static standard library on Android - and then maybe this crate shouldn't even specify any default so that the `CXXSTDLIB` variable can be used instead.

Zoxc:
I think my preferred option would be to have `cc` look for `-static-libstdc++` in `CXXFLAGS` and statically link if its not present. Ideally it would ask the C++ compiler to parse `CXXFLAGS` and give the required libraries.

rib:
The `cc` crate is effectively a Rust api-based C/C++ compiler front end for rust build scripts to use. To a large extent, I think the benefit of the crate is that you don't have to deal with the C/C++ compiler as a command line program and for most things also not have to handle platform specific compiler details or deal with environment variables.

The lack of nuance with environment variables like CXXFLAGS can make them pretty fragile if they affect everything in a build, but then I suppose that's also why you're thinking it could be a good place.

PR #19, opened 2022-08-25
Allow access to unicode chars for key events

Zoxc:
This gives some basic character events. `bindgen` should probably be run on this PR, but I don't know the magic invocation ;)

### DIFF
--- a/android-activity/game-activity-csrc/game-activity/GameActivity.cpp
+++ b/android-activity/game-activity-csrc/game-activity/GameActivity.cpp
@@ -1093,6 +1093,7 @@ static struct {
     jmethodID getRepeatCount;
     jmethodID getKeyCode;
     jmethodID getScanCode;
+    jmethodID getUnicodeChar;
 } gKeyEventClassInfo;
 
 extern "C" void GameActivityKeyEvent_fromJava(JNIEnv *env, jobject keyEvent,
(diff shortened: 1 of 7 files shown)


rib:
cool, yeah, this seems reasonable.

@Hoodad was also asking for something like this for Winit I think, for use with `ReceivedChar`.

There's also another version of `getUnicodeChar` that takes a modifiers argument but I'm not sure there's a practical way to expose that here. I suppose apps should ideally be using the IME interface eventually if they want proper unicode support.

For bindgen, you should be able to run the script as `./generate-bindings.sh` with `ANDROID_NDK_ROOT` in your environment (e.g. using Git Bash on Windows) it doesn't need any arguments. I can run though if you can't get it to run.

Zoxc:
I ran in to "missing" headers errors with `bindgen`. Not sure what's going on.

It also seems like only ASCII characters works with my keyboard. You can test [my APK](https://github.com/Zoxc/crusader/releases/tag/v0.0.7-testing) (I got CI builds working) which uses this PR + the winit PR.

rib:
> It also seems like only ASCII characters works with my keyboard. You can test [my APK](https://github.com/Zoxc/crusader/releases/tag/v0.0.7-testing) (I got CI builds working) which uses this PR + the winit PR.

right, I wouldn't be surprised if this only gives limited support for unicode characters yep - I think if you want proper unicode text handling that'll required getting IME based input hooked up.

I suppose I would have expected it to handle a least a bit more than ascii though :)

if it's really this limited though maybe we shouldn't land this.. this has a few trade offs:
1. it means every input event is now slightly larger, with an extra jni request to complete and it maybe doesn't provide any practically usable data
2. it creates a small inconsistency for the KeyEvent between NativeActivity and GameActivity. It's also notable that `getUnicodeChar` is by no means a new API ("Added in API level 1") but it's not even exposed by the [NDK `InputQueue` API](https://developer.android.com/ndk/reference/group/input) - which is maybe another indication that this API isn't really practically usefull.

Issue #18, opened 2022-08-21
Expose Rust API for input method handling via the GameTextInput library


rib:
For Android applications to be able to utilize soft keyboard UIs they need to be able to support the Android input methods protocol which the [GameTextInput](https://developer.android.com/games/agdk/game-activity/use-text-input) library helps with.

For reference this is the Winit IME API that was recently added that we should aim to be able to support: https://github.com/rust-windowing/winit/pull/2243 (and the original tracking issue: https://github.com/rust-windowing/winit/issues/1497)

Also see this Winit issue about supporting soft keyboards on Android: https://github.com/rust-windowing/winit/issues/2305

And here's an initial summary of what I found when considering supporting Winit's IME API on Android: https://github.com/rust-windowing/winit/issues/1497#issuecomment-1174053518

Zoxc:
I've read up on things a bit now and it doesn't look like `GameTextInput` is compatible with Winit's IME model. It expects the application's text buffer state to be accessible to read and write while in Winit's IME model text flows from the IME into the application.

A temporary hack would be to drain `GameTextInput`'s state for characters and generate `ReceivedCharacter` from these to get some basic support.

rib:
aah, okey, yeah interesting point.

Winit's IME abstraction is quite new, and was initially just aimed at supporting traditional desktop IME interfaces so it's not that surprising that it's not a good fit - though I hadn't realised that aspect about needing to be able to read back text entry state.

I would guess that we would be able to make additions to the Winit IME abstraction - so maybe it's something that we should first just get working in a branch - building on the android-activity-0.27 backend branch and hopefully it's possible to figure out a minimal addition that could support Android - and then whatever we add can probably just be stubbed out / ignored for platforms that don't need it.

I am a bit curious about this though - I would have thought other IMEs would also want to be able to sync with the contents of a text entry - unless they maybe only support appending text. I wonder if other platforms could maybe benefit from a way to access application text buffer state but they maybe didn't strictly need it.

rib:
For example with Wayland there is a [set_surrounding_text](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:request:set_surrounding_text) protocol that can inform an IME about the existing contents of a text entry  and that also ties into a related [delete_surrounding_text](https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:event:delete_surrounding_text) event that the IME can send for being able to delete characters from the original text.

rib:
[...]

PR #24, opened 2022-09-12
Input method (soft keyboard) support

rib:
Add input method support based on the GameTextInput library (only supported with `game-activity` currently)

This is based on discussion for #18 and [here](https://github.com/rib/android-activity/pull/21)

### DIFF
--- a/android-activity/src/game_activity/mod.rs
+++ b/android-activity/src/game_activity/mod.rs
@@ -32,6 +32,7 @@ use crate::{
 mod ffi;
 
 pub mod input;
+use crate::input::{TextInputState, TextSpan};
 use input::{Axis, InputEvent, KeyEvent, MotionEvent};
 
 // The only time it's safe to update the android_app->savedState pointer is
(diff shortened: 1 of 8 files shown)


rib:
Trying to test this, it doesn't currently seem to be working as intended.

I tried modifying agdk-mainloop so that it handles touches to the top-left of the screen by showing (toggling) the onscreen keyboard and printing input method state updates.

I see an input method 'commit' for every key press for some reason and the attempt to initialize the text state with some "Lorem ipsum" text and a selection didn't seem to work.

Once the keyboard is visible then other input handling seems to somehow fall apart, like touch events are being grabbed.

I may just look at landing the APIs for showing/hiding the keyboard first and a change that ensures that `android-activity` owns the (extensible, `'#[non_exhastive]`) `InputEvent` type so then it will hopefully be possible to get this working afterwards without needing a semver bump.

rib:
I ended up making incremental PRs for the show/hide_soft_input APIs and `InputEvent` enum as above so this updated PR just focuses on exposing a `TextEvent` and APIs for getting/setting the input method state.

android-activity/src/input.rs
```diff
@@ -0,0 +1,25 @@
...
+/// An empty span or cursor position is specified with `Some(start) == Some(end)`.
+///
+/// An undefined span is specified with start = end = `None`.
+#[derive(Debug, Clone, Copy)]
+pub struct TextSpan {
+    /// The start of the span (inclusive)
+    pub start: Option<usize>,
+
+    /// The end of the span (exclusive)
+    pub end: Option<usize>,
```
lucasmerlin, line 13: During my testing, leaving end None and providing only start, caused a android exception, so I think having start and end just be usize and then Option<TextSpan> in TextInputState might be better?

rib: I think I originally based this on these docs:

```
/**
 * This struct holds a span within a region of text from start (inclusive) to
 * end (exclusive). An empty span or cursor position is specified with
 * start==end. An undefined span is specified with start = end = SPAN_UNDEFINED.
 */
typedef struct GameTextInputSpan {
    /** The start of the region (inclusive). */
    int32_t start;
    /** The end of the region (exclusive). */
    int32_t end;
} GameTextInputSpan;
```

I think this issue is exactly what I got stuck on last time I was investigating this but didn't fully dig into the details.

Following the flow of data here...

- This data gets passed to `GameActivity_setTextInputState` with `None` mapped to `-1` based on the above comment.

- `GameActivity_setTextInputState` passes this to `GameTextInput_setState` via a `CMD_SET_SOFT_INPUT_STATE` callback.

`GameTextInput_setState` calld `GameTextInput::setState` which does a JNI call to `gametextinput.InputConnection.setState` in Java (https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/refs/heads/main/game-text-input/src/main/java/com/google/androidgamesdk/gametextinput/InputConnection.java#196)

`setState` is a utility that then effectively calls `InputConnection.setSelection` and `InputConnection.setComposingRegion` which are standard APIs documented here: https://developer.android.com/reference/android/view/inputmethod/InputConnection#setComposingRegion(int,%20int)

It looks like `InputConnection` is expected to clamp indices to being in-bounds and also not be fussy about wither start comes after end. At least the docs for `setComposingRegions` say "The passed indices are clipped to the contents bounds" which suggests it should be ok to pass `None/-1` for the start/end indices.

Maybe the java GameTextInput code isn't validating its inputs properly.

E.g. looking at https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/refs/heads/main/game-text-input/src/main/java/com/google/androidgamesdk/gametextinput/GameTextInput.java#69

It look like it should also be considering indices that are out of bounds because they are negative.

Or maybe only `setComposingRegions` is designed to clamp/clip the indices and you're expected to keep them in-bounds for `setSelection`.

Conceptually it looks like there is more of a special case for `setComposingRegion` which specifically recognises a start index of `-1` and will map that to `removeComposingRegion`.

Maybe the comments that were written for `GameTextInputSpan` about using `SPAN_UNDEFINED` only really apply for the composing region.

That would maybe make sense considering that the selection becomes a cursor if the start/end are the same but you can't completely remove the selection _and_ the cursor too.

As you suggested - moving the Options up a level probably makes sense.

I'm currently think that maybe only the `compose_region` should be optional though.

For the selection then if it was set to `None` then we'd have to arbitrarily choose to either put the cursor at the start of end I guess but it seems like that should be explicitly decided by the app probably.

rib:
Okey, I've rebased this on GameActivity 2.0.2 and I've made it so TextSpan just has `usize` start/end values but the `composite_region` is now an `Option`.

@lucasmerlin it'd be great if you're able to test this based on your Winit / egui changes.

android-activity/src/game_activity/mod.rs
```diff
@@ -424,6 +522,15 @@ impl AndroidAppInner {
...
         }
+
+        unsafe {
+            let app_ptr = self.native_app.as_ptr();
+            if (*app_ptr).textInputState != 0 {
+                let state = self.text_input_state();
+                callback(&InputEvent::TextEvent(state));
+                (*app_ptr).textInputState = 0;
+            }
+        }
```
lucasmerlin, line 533: For some reason, when I try the new changes, the text input state isn't updated while I'm typing. Only when some other event is triggered I can see the updated text in my app. 
Moving this block to the beginning of the function (so it's before the return in line 512) fixes this.
I guess the textInputState is independent of the input buffers, so the early return breaks this. 

I'm not sure why this worked previously though?

---

I figured it out lol, it worked previously because it was above the return and was moved down

android-activity/src/input.rs
lucasmerlin: My updated egui code looks a bit cleaner now, so I think this makes sense. I'm not 100% sure though on whether selection should be optional or not.

android-activity/src/input.rs
```diff
@@ -81,3 +81,24 @@ impl From<Source> for Class {
         source.into()
     }
 }
+/// This struct holds a span within a region of text from `start` (inclusive) to
+/// `end` (exclusive).
+///
+/// An empty span or cursor position is specified with `Some(start) == Some(end)`.
```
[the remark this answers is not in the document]
rib: oh, I removed the line below about specifying None but forgot to remove the `Some()` wrappers from this line.

I guess we still want the line itself that notes how an empty span can be represented.

android-activity/src/game_activity/mod.rs
rib: Ah, right, it looks like my old branch had the same bug which is that `input_events` can return early if it sees that `android_app_swap_input_buffers` returns an empty buffer.

I didn't see what you mean about this previously being after the return - did you maybe modify my branch for testing? The older code I'm looking at has this chunk at the end of `input_events`

lucasmerlin: Oh, you're right, it was actually a change I did. I guess I did ran into this problem before but forgot about it.

rib:
>  I'm not 100% sure though on whether selection should be optional or not.

Yeah I'm not 100% sure about the handling of the optional spans.

Looking at the docs for `InputConnection.setSelection` and `.setComposingRegion` (which are basically what these spans are for) then technically neither of these APIs support a special case of an unset/undefined regions (they both just recognise when the `start == end`)

- https://developer.android.com/reference/android/view/inputmethod/InputConnection#setSelection(int,%20int)
- https://developer.android.com/reference/android/view/inputmethod/InputConnection#setComposingRegion(int,%20int)

so from that pov it could also be reasonable that the `composing_region` wouldn't be optional either but because the GameActivity-specific subclass implement's a special case that recognises `start == -1` that's pretty much the only reason I can currently see for supporting an optional `composing_region`.

Tbh it seems odd to me that the GameActivity implementation of InputConnection has this special case which isn't documented in the context of `GameTextInputState` or `GameTextInput_setState`.

Passing None for the composing_region will result in `editable.removeSpan()` for the private / dummy editable that the GameActivity InputConnection has internally.

Considering that it's really just a placeholder editable it doesn't matter much I expect that this will remove the span.

I'm fairly sure that the Java implementation should also remove the span in case an empty region is specified, considering the SDK docs for `setComposingRegion`

> If the resulting region is zero-sized, no region is marked and the effect is the same as that of calling [finishComposingText()](https://developer.android.com/reference/android/view/inputmethod/InputConnection#finishComposingText()).

So actually maybe the right thing to do in android-activity is to recognise zero-sized composing regions ourselves and then normalize that into a `-1,-1` region, which will result in `editable.removeSpan()`, which looks like a better interpretation of what the connection should be doing anyway.

This would mean we can remove the `Option` for the composing_region too.

rib:
Although it could make sense for the composing_region to be non-optional when setting the region, it would be awkward when querying the state because it needs to be optional when reading the state since the start/end will be -1 if no region has been set yet or if it's is cleared.

It would probably still be best to represent an unset composing region with `None` instead of picking an arbitrary `start==end` pair.

rib:
oh, oops, sorry I just re-requested review @lucasmerlin  but I forgot I was going to change it so we check for an empty compose region in Rust and normalize that as -1,-1 - will do that now

rib:
> oh, oops, sorry I just re-requested review @lucasmerlin but I forgot I was going to change it so we check for an empty compose region in Rust and normalize that as -1,-1 - will do that now

okey, done

I'm hoping that if you're perhaps able to smoke test this with your winit/egui branch @lucasmerlin then we can perhaps land this so we can potentially look at getting the corresponding winit changes into 0.29 if lucky

lucasmerlin (approved):
Works perfectly now! 
Only missing thing now is some way to set the keyboard flags, but I think that can be done in a separate PR. 
I currently use these for Autocorrect + Sentence capitalization:
```
            ffi::GameActivity_setImeEditorInfo(activity, 1 | 16384, 0, 0);
```

rib:
Cool, thanks for checking, and yeah we can look at adding API for keyboard flags via a follow up PR

lucasmerlin:
I've opened draft PRs to implement this for [winit](https://github.com/rust-windowing/winit/pull/2993) and [egui](https://github.com/emilk/egui/pull/3196)

PR #29, opened 2022-09-18
update na-winit-wgpu example

msiglreith:
- update shader to match syntax changes
- use `native-activity` instead of `game-activity`


rib:
Ah, I could have sworn I made a patch at some point to re-sync this shader, oops.

thanks for fixing this - it should probably built in CI too at some point

PR #39, opened 2022-11-10
Release 0.4

rib:


rib:
Just for reference here; Clippy highlighted a good issue that's lead me to want to try and address two things before finalizing this release:

1) MotionEvents with the GameActivity backend are kinda huge at the moment because they actually copy the full event state out of the circular buffer that's used to collect them (which is over-sized based on the largest number of axis supported) - whereas they should effectively just point into the buffer with a suitable lifetime (which should be doable since we only expose input events via a reference with a lifetime to the `input_events` callback.

An annoyance here will be that we can't use GATs atm to have a lending iterator of the input buffer since our MSRV is 1.60 but since the iteration is all private there's no requirement to work in terms of an `Iterator` like we do currently.

2) MotionEvents and KeyEvents with the NativeActivity backend simply expose the `ndk` crate types which look they probably aren't as safe as I had assumed since they are just `NonNull` pointer wrappers but with no associated lifetime. I think that means that there would be nothing stopping code copying `MotionEvent` or `KeyEvent` structs and potentially keeping ahold of them until the underlying pointer may become invalid.

Again I think since android-activity only exposes input events by reference within `input_event` callbacks we can simply add our own wrappers for MotionEvent and KeyEvent with suitable lifetime parameters that will block code from keeping these pointers too long. I guess I'll needing to mirror the ndk events API explicitly (similar to what is already done for GameActivity) to avoid publicly exposing the ndk types directly. (as a side-effect this should result in better compatibility since it will be more consistent with the GameActivity backend)

Fairly sure these can both be addressed without (practically) changing the public API, but since it will technically be messing with the types used to expose events it makes sense to address them before the 0.4 release.

Issue #60, opened 2023-02-13
Expose `JavaVM` and `Activity` reference that can be used to make JNI calls


rib:
So that we can port crates away from the `ndk-context` crate we should expose the `JavaVM` and Android `Activity` reference needed for applications to use JNI to also call Java/Kotlin APIs from their native code.

One reason why we want to deprecate the `ndk-context` crate is that it relies on global static state for sharing the `Activity` reference which makes it impossible to support applications that may have more than one native activity running in the same process.

MarijnS95:
@rib is the intent to have "random" crates (`app_dirs2` comes to mind) take an `AndroidApp` to extract these and suddenly be locked in to `android-activity` version cadence, or rather pass the "untyped" `*mut c_void` directly (via the _user of_ both crates) in a builder pattern as suggested in https://github.com/rust-mobile/android-activity/issues/58#issuecomment-1398468102?

rib:
> @rib is the intent to have "random" crates (`app_dirs2` comes to mind) take an `AndroidApp` to extract these and suddenly be locked in to `android-activity` version cadence, or rather pass the "untyped" `*mut c_void` directly (via the _user of_ both crates) in a builder pattern as suggested in [#58 (comment)](https://github.com/rust-mobile/android-activity/issues/58#issuecomment-1398468102)?

Ideally I think crates like `app_dirs2` probably shouldn't depend on the `android-activity` crate (that could actually create a versioning nightmare, similar to how lots of crates used to depend on `ndk-glue` before `ndk-context` existed - since there can only be one version of `android-activity` linked into an application)

Instead it should either be possible to initialize `app_dirs2` based on the raw pointers or based on the `jni` types `&JavaVM` and `&JObject` (ideally both, to give users flexibility with where they acquire the JNI pointers from)

This way it's the application that glues these things together and they can decide whether they will e.g. use `android-activity` or maybe just `JNI_OnLoad` or `ndk-context` etc

MarijnS95:
> since there can only be one version of `android-activity` linked into an application

At least explicitly passing `app: AndroidApp` will result in a compiler error as the two types come from different versions of the same crate (or the exported symbol will clash) instead of vague runtime errors we used to get before introducing `ndk-context`. But yes, you wouldn't want every crate needing the `VM`/`Activity` depend on the "entrypoint" `android-activity` crate.

> ideally both, to give users flexibility with where they acquire the JNI pointers from

[...]

PR #61, opened 2023-02-13
Add AndroidApp::vm_as_ptr() and ::activity_as_ptr() APIs

rib:
This enables applications to make JNI calls without needing the `ndk-context` crate - which we would like to deprecate.

For now this just exposes the raw pointers, and makes no assumptions about what crates an application will use to make JNI calls - which means that some amount of `unsafe` code will be required when passing these to other crates.

The added `_as_ptr()` suffix makes this clear, and also means we can consider adding similar (suffix-free) getters later that would return `jni` crate types that could reduce the need for unsafe code before making JNI calls.

Fixes: #60

### DIFF
--- a/android-activity/src/game_activity/mod.rs
+++ b/android-activity/src/game_activity/mod.rs
@@ -12,6 +12,7 @@ use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use std::{ptr, thread};
 
+use libc::c_void;
 use log::{error, trace, Level};
 
 use jni_sys::*;
@@ -155,6 +156,16 @@ pub struct AndroidAppInner {
 }
 
 impl AndroidAppInner {
+    pub fn vm_as_ptr(&self) -> *mut c_void {
+        let app_ptr = self.native_app.as_ptr();
+        unsafe { (*(*app_ptr).activity).vm as _ }
+    }
+
+    pub fn activity_as_ptr(&self) -> *mut c_void {
+        let app_ptr = self.native_app.as_ptr();
+        unsafe { (*(*app_ptr).activity).javaGameActivity as _ }
+    }
+
     pub fn native_window(&self) -> Option<NativeWindow> {
         self.native_window.read().unwrap().clone()
     }
--- a/android-activity/src/lib.rs
+++ b/android-activity/src/lib.rs
@@ -62,6 +62,7 @@ use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::Duration;
 
+use libc::c_void;
 use ndk::asset::AssetManager;
 use ndk::native_window::NativeWindow;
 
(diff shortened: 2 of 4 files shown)


android-activity/src/lib.rs
```diff
@@ -472,6 +473,50 @@ impl AndroidApp {
         self.inner.read().unwrap().native_window()
     }
 
+    /// Returns a pointer to the Java Virtual Machine, for making JNI calls
+    ///
+    /// This returns a pointer to the Java Virtual Machine which can be used with the
+    /// [jni] crate (or similar crates) to make JNI
+    /// calls that bridge between native Rust code and Java/Kotlin code running within
+    /// the JVM.
```
MarijnS95, line 481: Wrapping here seems a bit odd, and should `jni` be between backticks as it's a crate name?

android-activity/src/lib.rs
```diff
@@ -472,6 +473,50 @@ impl AndroidApp {
...
+        self.inner.read().unwrap().vm_as_ptr()
+    }
+
+    /// Returns a JNI object reference for this application's JVM `Activity` as a pointer
+    ///
+    /// If you use the `jni` crate you can wrap this as an object reference via:
+    /// ```ignore
+    /// # use jni::objects::JObject;
+    /// # let app: AndroidApp = todo!();
+    /// let activity = unsafe { JObject::from_raw(app.android_activity_as_ptr()) };
```
MarijnS95, line 502:
```suggestion
    /// let activity = unsafe { JObject::from_raw(app.activity_as_ptr()) };
```

android-activity/src/game_activity/mod.rs
```diff
@@ -155,6 +156,20 @@ pub struct AndroidAppInner {
 }
 
 impl AndroidAppInner {
+    pub fn vm_as_ptr(&self) -> *mut c_void {
+        unsafe {
+            let app_ptr = self.native_app.as_ptr();
```
MarijnS95, line 161: `as_ptr()` isn't `unsafe` so this line can stay outside the `unsafe {}` block?

MarijnS95 (approved):

MarijnS95:
> @MarijnS95 hopefully this looks like a reasonable initial solution for exposing the JNI pointers we need without needing the `ndk-context` crate?

I think, as long as there is a builder pattern available to explicitly pass these values from `fn main()` to crates (like `app_dirs2`) that were previously consuming `ndk-context`; and users are aware that they have to glue these pieces together.

rib:
Thanks for looking over the patch. I've applied the suggested changes.

> > @MarijnS95 hopefully this looks like a reasonable initial solution for exposing the JNI pointers we need without needing the `ndk-context` crate?
> 
> I think, as long as there is a builder pattern available to explicitly pass these values from `fn main()` to crates (like `app_dirs2`) that were previously consuming `ndk-context`; and users are aware that they have to glue these pieces together.

Yeah - probably my main concern with this direction atm is that the ergonomics aren't going to be great. These APIs that accept raw pointers will need to be `unsafe` which then leads you to put all the building inside an `unsafe` block or else you need to split the building pattern up, which defeats the benefit of the pattern.

Regardless of exactly how the dots are joined though I think it still makes sense that there is at least some way to access the JNI pointers directly from an `AndroidApp` now, so we have options for figuring out how crates like `app_dirs2` can be initialized without depending on `ndk-context`.

PR #68, opened 2023-03-20
Add panic guards to extern "C" functions

notgull:
Prevents undefined behavior in the event that any functions panic.

### DIFF
--- a/android-activity/src/native_activity/glue.rs
+++ b/android-activity/src/native_activity/glue.rs
@@ -15,8 +15,7 @@ use std::{
 use log::Level;
 use ndk::{configuration::Configuration, input_queue::InputQueue, native_window::NativeWindow};
 
-use crate::util::android_log;
-use crate::ConfigurationRef;
+use crate::{util::abort_on_panic, util::android_log, ConfigurationRef};
 
 use super::{AndroidApp, Rect};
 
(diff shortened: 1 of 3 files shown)


MarijnS95:
Previously folks used panic_unwind to also print the error/backtrace, is this the desired way now?

notgull:
I'd also be fine with `panic_unwind`, it's just that from the codebase I get the impression that a panic here is likely a critical bug that can't be recovered from

rib:
thanks, having something to map panics in extern C functions to an `abort` sounds good. (at the very least we shouldn't be letting panics unwind across any Java or C FFI boundaries)

I didn't follow what extra protection was provided by the `AbortOnPanic` struct with the `Drop` handler when the code uses `catch_unwind` to catch any panic+unwind and abort? The `AbortOnPanic` stuff seemed redundant?

Adding the `catch_unwind` around `android_main` is probably the most important thing here that we should have done before (at least for most of the other ones they would hopefully only panic due to an internal bug, but a panic from the application will be a common occurrence)

notgull:
> I didn't follow what extra protection was provided by the `AbortOnPanic` struct with the `Drop` handler when the code uses `catch_unwind` to catch any panic+unwind and abort? The `AbortOnPanic` stuff seemed redundant?

This protects undefined behavior from occurring if the panic-handling code panics.

MarijnS95:
> I'd also be fine with `panic_unwind`, it's just that from the codebase I get the impression that a panic here is likely a critical bug that can't be recovered from.

Yeah, so keep the `abort()` but still print a friendly message which won't happen with the original implementation? It seems you're doing that now though.

android-activity/src/native_activity/glue.rs
```diff
@@ -738,92 +765,135 @@ extern "C" fn ANativeActivity_onCreate(
...
+                    android_log(Level::Error, rust_panic, &msg);
+                }
+            } else {
+                let unknown_panic =
+                    unsafe { CStr::from_bytes_with_nul_unchecked(b"UnknownPanic\0") };
+                android_log(Level::Error, unknown_panic, &CString::new("").unwrap());
+            }
+
+            std::process::abort();
+        });
```
MarijnS95, line 895: Can we dedup this between the two backends?

MarijnS95: (Same for the `android_log` function, that's common code)

rib:
> Prevents undefined behavior in the event that any functions panic.

Ah, right, I see. If we're worried that the catch code might panic though, can we consider just using `catch_unwind` for both cases for readability. The intent for the Drop code wasnt super obvious to me at least and I think the closure scope for a second `catch_unwind` might have made it more obvious that we want to also handle secondary panics too.

I'm also wondering if we should specifically catch around the application's `android_main` to distinguish how we deal with an internal panic vs an application panic.

android-activity/src/util.rs
```diff
@@ -13,3 +18,57 @@ pub fn try_get_path_from_ptr(path: *const c_char) -> Option<std::path::PathBuf>
...
+                    android_log(Level::Error, rust_panic, &msg);
+                }
+            } else if let Ok(panic) = panic.downcast::<&str>() {
+                if let Ok(msg) = CString::new(*panic) {
+                    android_log(Level::Error, rust_panic, &msg);
+                }
+            } else {
+                let unknown_panic =
+                    unsafe { CStr::from_bytes_with_nul_unchecked(b"UnknownPanic\0") };
+                android_log(Level::Error, unknown_panic, &CString::new("").unwrap());
```
MarijnS95, line 66: `CStr::from_bytes_with_nul_unchecked(b"\0")` for the `msg`?

rib:
For reference here, I rebased your changes @notgull and also updated the implementation of `abort_on_panic` to use `catch_panic` for the extra-paranoid check in case we panic while logging the first panic.

I've also put a `TODO` comment in there to note that actually we probably shouldn't be aborting at all.

I recently raised it as an issue in Winit recently that is problematic that Winit's `run` API would exit the whole process and the same thing applies here I think. An Android Activity doesn't own the whole process because there could be other `Services`, or even other `Activity`s running in other threads which we shouldn't be nuking if we've safely caught a panic.

At some point we should probably throw appropriate Java exceptions via JNI instead.

I haven't yet looked at putting a more precise catch around the application's `android_main` which I think would make sense here too (so we can handle app panics differently to internal panics).

android-activity/src/util.rs
```diff
@@ -13,3 +17,56 @@ pub fn try_get_path_from_ptr(path: *const c_char) -> Option<std::path::PathBuf>
...
+        Level::Trace => ndk_sys::android_LogPriority::ANDROID_LOG_VERBOSE,
+    };
+    unsafe {
+        ndk_sys::__android_log_write(prio.0 as libc::c_int, tag.as_ptr(), msg.as_ptr());
+    }
+}
+
+/// Run a closure and abort the program if it panics.
+///
+/// This is generally used to ensure Rust callbacks won't unwind past the JNI boundary, which would lead
```
MarijnS95, line 36:
```suggestion
/// This is generally used to ensure Rust callbacks won't unwind past the JNI boundary, which leads
```

Perhaps?

android-activity/src/util.rs
```diff
@@ -13,3 +17,56 @@ pub fn try_get_path_from_ptr(path: *const c_char) -> Option<std::path::PathBuf>
...
+/// to undefined behaviour.
+///
+/// TODO(rib): throw a Java exception instead of aborting. An Android Activity does not necessarily
+/// own the entire process because other application Services (or even Activities) may run in
+/// threads within the same process, and so we're tearing down too much by aborting the process.
+pub(crate) fn abort_on_panic<R>(f: impl FnOnce() -> R) -> R {
```
MarijnS95, line 42: This seems like a thing to solve soon or make an issue for (panicking only affects the current thread, abort kills the whole process).

android-activity/src/util.rs
```diff
@@ -13,3 +17,56 @@ pub fn try_get_path_from_ptr(path: *const c_char) -> Option<std::path::PathBuf>
...
+                let unknown_panic =
+                    unsafe { CStr::from_bytes_with_nul_unchecked(b"UnknownPanic\0") };
+                android_log(Level::Error, unknown_panic, unsafe {
+                    CStr::from_bytes_with_nul_unchecked(b"\0")
+                });
+            }
+
+            std::process::abort();
+        }))
+        .unwrap_or_else(|_panic| std::process::abort())
```
MarijnS95, line 70: Can we just call this, once, outside the `catch_unwind` instead of having a second `unwrap_or_else`?

MarijnS95 (approved):
Few suggestions, but generally looks okay.

android-activity/src/util.rs
rib: Yeah, we should aim to iterate on this for sure.

In the short term I'd be ok with landing merging this since I at least figure that this is better than the previous situation where a panic could cross the JNI boundary which would be undefined behaviour.

rib: just created this issue to track: https://github.com/rust-mobile/android-activity/issues/80

PR #76, opened 2023-04-22
Add MSRV policy to README and bump `rust_version` to 0.64, due to `ndk` -> `raw_window_handle` dependency

rib:
This also updates the CI to test with `cargo-ndk` 2, which compiles with rustc 1.64 (whereas v3 requires 1.68)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,9 @@ jobs:
           i686-linux-android
 
       - name: Install cargo-ndk
-        run: cargo install cargo-ndk
+        # We're currently sticking with cargo-ndk 2, until we bump our
+        # MSRV to 1.68+
+        run: cargo install cargo-ndk --version "^2"
 
       - name: Setup Java
         uses: actions/setup-java@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         # We need to support the same MSRV as Winit which we are currently
         # assuming will be bumped to 1.60.0 soon:
```
MarijnS95, line 20: Update stale comment?

android-activity/Cargo.toml
```diff
@@ -9,6 +9,7 @@ repository = "https://github.com/rust-mobile/android-activity"
 documentation = "https://docs.rs/android-activity"
 description = "Glue for building Rust applications on Android with NativeActivity or GameActivity"
 license = "MIT OR Apache-2.0"
+rust_version = "0.64"
```
MarijnS95, line 12: That's old, did a `0.64` release even exist?

```suggestion
rust_version = "1.64"
```

MarijnS95:
Title also has `0.64` but I think you meant `1.64` :)

Issue #67, opened 2023-03-16
Call `ANativeActivity_finish()` or `GameActivity_finish()` when `android_main` returns


rib:
Additionally, expose an `AndroidApp::finish()` API that can be used to explicitly call the `Activity::finish()` method which will lead to the activity being stopped and destroyed (with a corresponding `Destroy` event)

MarijnS95:
(Edited your message to include the full referenced doc-comment lines in-line :slightly_smiling_face:)

Indeed, we may have been wrong in assuming for and allowing the thread-spawn. Good to have the back-link to check our ideas, but at this point I'd say to go for it and call `finish()` after returning (and document this in the changelog), and see if anyone gets broken by it which seems unlikely given the above rules :)

PR #81, opened 2023-04-27
Call Activity.finish() when android_main returns

rib:
Calling Activity.finish() is what ensures the Activity will get gracefully destroyed, including calling the Activity's onDestroy method.

Fixes: #67

### DIFF
--- a/android-activity/src/native_activity/glue.rs
+++ b/android-activity/src/native_activity/glue.rs
@@ -8,6 +8,7 @@ use std::{
     io::{BufRead, BufReader},
     ops::Deref,
     os::unix::prelude::{FromRawFd, RawFd},
+    panic::catch_unwind,
     ptr::{self, NonNull},
     sync::{Arc, Condvar, Mutex, Weak},
 };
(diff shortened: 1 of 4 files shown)


android-activity/src/native_activity/glue.rs
```diff
@@ -874,15 +879,26 @@ extern "C" fn ANativeActivity_onCreate(
...
-                // springboard (e.g. call an Activity subclass method that calls a jni native
-                // method that then just calls android_main()) that would make sure there was
-                // a Java frame at the base of our call stack which would then be recognised
-                // when calling FindClass to lookup a suitable classLoader, instead of
-                // defaulting to the system loader. Without this then it's difficult for native
-                // code to look up non-standard Java classes.
-                android_main(app);
+                // We want to specifically catch any panic from the application's android_main
+                // so we can finish + destroy the Activity gracefully via the JVM
+                catch_unwind(|| {
```
MarijnS95, line 884: Personally a fan of keeping `unsafe` blocks as small as possible, now we're wrapping quite a lot of "normal" code in it.

MarijnS95 (approved):
Nice, especially a good idea to also invoke all this cleanup when `android_main()` returns via panicking/unwinding :)

PR #84, opened 2023-05-23
game_activity: Fix `pointer_index()` always returning `0`

yunsash:
Given `AMOTION_EVENT_ACTION_MASK` is 0xff and `AMOTION_EVENT_ACTION_POINTER_INDEX_MASK` is 0xff00, this function always returns 0. Looks like ANDing with `AMOTION_EVENT_ACTION_MASK` should be removed.


rib:
Thanks, yeah this looks wrong. I doubt I was planning to validate the action and probably just made a mistake at the time, oops.

PR #85, opened 2023-05-28
Bump `bitflags` to v2

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented by the macro is now different.

I was a bit liberal with deriving `std` traits on public types, which mimics the behavior of `bitflags` v1, let me know if we don't need all these traits. Potentially we could at least drop `PartialOrd` and `Ord`.

[Replaces https://github.com/zakarumych/gpu-alloc/pull/70](https://github.com/rust-mobile/android-activity/pull/83).

### DIFF
--- a/android-activity/src/input.rs
+++ b/android-activity/src/input.rs
@@ -36,6 +36,7 @@ pub enum Source {
 }
 
 bitflags! {
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
     struct SourceFlags: u32 {
         const CLASS_MASK = 0x000000ff;
 
--- a/android-activity/src/lib.rs
+++ b/android-activity/src/lib.rs
@@ -275,6 +275,7 @@ pub use activity_impl::AndroidAppWaker;
 bitflags! {
     /// Flags for [`AndroidApp::set_window_flags`]
     /// as per the [android.view.WindowManager.LayoutParams Java API](https://developer.android.com/reference/android/view/WindowManager.LayoutParams)
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
     pub struct WindowManagerFlags: u32 {
         /// As long as this window is visible to the user, allow the lock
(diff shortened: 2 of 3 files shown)


android-activity/src/lib.rs
```diff
@@ -275,6 +275,7 @@ pub use activity_impl::AndroidAppWaker;
 bitflags! {
     /// Flags for [`AndroidApp::set_window_flags`]
     /// as per the [android.view.WindowManager.LayoutParams Java API](https://developer.android.com/reference/android/view/WindowManager.LayoutParams)
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
rib, line 278: same as above, these flags don't have a natural ordering so we probably shouldn't derive `PartialOrd/Ord`

android-activity/src/input.rs
```diff
@@ -36,6 +36,7 @@ pub enum Source {
 }
 
 bitflags! {
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
rib, line 39: I guess we shouldn't include `PartialOrd/Ord` here

rib (changes requested):
Thanks, looks good except for deriving PartialOrd/Ord for the flags which I'm not sure we want.

Thanks if you update the branch with that tweak but no worries if not I can tweak before merging.

android-activity/src/input.rs
MarijnS95: It is quite a nice feat that `bitflags` now doesn't give you some of these "nonsense" default derives :)

rib: Just removed these and rebased the branch

daxpedda:
Apologies for my absence, thanks for getting this done!

PR #86, opened 2023-06-16
README: Add badges to CI, crates.io, docs.rs and show the MSRV

MarijnS95:
I was trying to quickly get to the documentation of this crate and had the GitHub page open... but there was no link on the front-page: let's fix that.


rib (approved):
Thanks!

PR #87, opened 2023-06-23
cargo: Fix `rust_version` -> `rust-version` property typo

MarijnS95:
Cargo complains:

    warning: android-activity/Cargo.toml: unused manifest key: package.rust_version

Solve this by replacing the underscore with a hyphen.


rib (approved):
Ah, good catch, thanks!

MarijnS95:
Wish cargo had a `-Werror` mode to guard against such things in CI :)

PR #89, opened 2023-06-25
Add some initial notes for updating game-activity glue

rib:
This digs up a few notes from when I first implemented that agkd-rust crate that later became android-activity.

Hopefully we can prune the set of patches that we're tracking the next time that we sync with a new AGDK GameActivity release.

TODO: It would be good to have improved workflow documentation for how we will pull in new AGDK releases.

TODO: We should also document some kind of policy for what kinds of changes will be considered semver breaking changes for the android-activity crate.


rib:
Closing this for now since the merge of https://github.com/rust-mobile/android-activity/pull/88 leaves a reasonable paper trail of the steps and patches involved in updating to the latest GameActivity release. All the patches are back-to-back currently so it's quite clear what changes we're maintaining on top of the upstream GameActivity glue code.

Issue #97, opened 2023-07-30
Fix deadlocks in game-activity backend if java callbacks are fired after native thread exits


rib:
See the discussion here for more context: https://github.com/rust-mobile/android-activity/pull/94#issuecomment-1652014915

Essentially the GameActivity backend doesn't have checks within its Java Activity callbacks that check whether the `android_app` has already been destroyed (e.g. due to a graceful exit by returning from `android_main`

Since `android-activity` will issue a GameActivity_finish when android_main returns that will result in `onPause -> onStop -> onDestroyed` lifecycled events in Java after the native thread for `android_main` has exit and then because the GameActivity callbacks don't check if the `android_app` is detroyed they deadlock waiting for the native thread to handle these lifecycle events.

PR #98, opened 2023-07-30
GameActivity PATH: fix deadlocks in java callbacks after app destroyed

rib:
This ensures that any java Activity callbacks take into account the possibility that the `android_app` may have already been marked destroyed if `android_main` has returned - and so they mustn't block and wait for a thread that is no longer running.

Fixes: #97

### DIFF
--- a/android-activity/build.rs
+++ b/android-activity/build.rs
@@ -1,6 +1,12 @@
 #![allow(dead_code)]
 
 fn build_glue_for_game_activity() {
+    for f in [
+        "GameActivity.h",
+        "GameActivity.cpp",
+    ] {
+        println!("cargo:rerun-if-changed=game-activity-csrc/game-activity/{f}");
+    }
     cc::Build::new()
         .cpp(true)
         .include("game-activity-csrc")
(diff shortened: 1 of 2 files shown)


rib:
I've at least tested this with the agdk-mainloop example modified to exit early and so I've seen that successfully call `onPause`, `onStop` and `onDestroy` after the `android_app` was marked destroyed.

I'm going to go ahead an land this since I'd like to move on to landing some of the bigger changes like GameActivity 2.0.2 before the next Winit beta release.

@ardocrat it'd be good if you can test this to confirm it fixes the deadlock you were hitting.

Feel free to comment here even if though I'm going to merge this optimistically now.

ardocrat:
Thanks, this case seems fixed now! After calling of `eframe::Frame::close` I can see `onDestroy()` passed.

But now I have problem with `winit`, seems like `EventLoop` is not finishing, I got an error after relaunch: `Creating EventLoop multiple times is not supported`, I am using `eframe = 0.22.0` and `winit = 0.28`.

My log:
```
21:50:33.404  D  loadNativeCode_native
21:50:33.406  V  Creating: 0x722f9271f0
21:50:33.406  V  Callbacks set: 0x722f927240
21:50:33.406  V  Launching android_app_entry in a thread
21:50:33.406  V  android_app_entry called
21:50:33.406  V  android_app = 0x7137dbf040
21:50:33.406  V  config = 0x717f939d10
21:50:33.406  V  activity = 0x722f9271f0
21:50:33.406  V  assetmanager = 0x721f93fcc0
21:50:33.406  V  Config: mcc=250 mnc=99 lang=en cnt=GB orien=1 touch=3 dens=560 keys=1 nav=1 keysHid=3 navHid=0 sdk=30 size=2 long=2 modetype=1 modenight=2
21:50:33.434  V  onStart_native
21:50:33.434  V  Start: 0x722f9271f0
21:50:33.442  I  thread '<unnamed>' panicked at 'Creating EventLoop multiple times is not supported.', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.28.6/src/event_loop.rs:116:13
21:50:33.442  I  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
21:50:33.443  E  Creating EventLoop multiple times is not supported.
21:50:33.443  D  write_work: cmd=1
21:50:33.443  V  android_app_destroy!
```

rib:
> I think I will kill process after onDestroy()

Just in case you might ever want to run other Android Services in the same process I'd try and avoid explicitly killing the process if you can avoid it.

The lifecycle for the activity is separate from the lifecycle of services that could be running in the same process.

On the Winit side of things then things should be a bit better in 0.29 hopefully, re: https://github.com/rust-windowing/winit/issues/2900

PR #110, opened 2023-08-15
native-activity: don't treat missing input queue as error

rib:
If an app tries to iterate input events while there's no input queue (e.g. before onStart) then just behave like there are no events available instead of returning an error.

This also reduces the logging level of some messages to reduce the verbosity of info logs.

### DIFF
--- a/android-activity/src/native_activity/mod.rs
+++ b/android-activity/src/native_activity/mod.rs
@@ -13,7 +13,7 @@ use log::{error, trace};
 use ndk::input_queue::InputQueue;
 use ndk::{asset::AssetManager, native_window::NativeWindow};
 
-use crate::error::{InternalAppError, InternalResult};
+use crate::error::InternalResult;
 use crate::input::{Axis, KeyCharacterMap, KeyCharacterMapBinding};
 use crate::input::{TextInputState, TextSpan};
 use crate::jni_utils::{self, CloneJavaVM};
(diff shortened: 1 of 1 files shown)


android-activity/src/native_activity/mod.rs
```diff
@@ -486,58 +485,63 @@ impl<'a> InputIteratorInner<'a> {
...
-                let event = match ndk_event {
-                    ndk::event::InputEvent::MotionEvent(e) => {
-                        input::InputEvent::MotionEvent(input::MotionEvent::new(e))
-                    }
-                    ndk::event::InputEvent::KeyEvent(e) => {
-                        input::InputEvent::KeyEvent(input::KeyEvent::new(e))
-                    }
-                };
+        // XXX: would use `let Some(queue) = &self.receiver.queue else { return false; }`
+        // to avoid rightward drift but we're stuck supporting Rust 1.64 for Winit currently
```
MarijnS95, line 489: You can achieve the very same with `let queue = if let Some(queue) = &self.receiver.queue { queue } else { return false; }` to get rid of rightward drift _now_, and simplify this statement when `let-else` is supported by MSRV.

Makes it easier to read the diff _without_ disabling whitespace changes.

rib: ah, yep, would be a nice compromise

MarijnS95 (approved):
Rest looks fine to me, nice to catch the `info!` logs before a release :)

MarijnS95 (approved):

PR #113, opened 2023-08-15
Upgrade to `ndk-sys 0.5.0-beta.0`, `ndk 0.8.0-beta.0`

MarijnS95:
~Depends on https://github.com/rust-mobile/ndk/pull/420.~


rib (approved):

examples/agdk-mainloop/Cargo.toml
```diff
@@ -9,8 +9,8 @@ edition = "2021"
 log = "0.4"
 android_logger = "0.11.0"
 android-activity = { path="../../android-activity", features = ["game-activity"] }
-ndk-sys = "0.4"
-ndk = "0.7"
+ndk-sys = "0.5.0-beta.0"
+ndk = "0.8.0-beta.0"
```
MarijnS95, line 13: @rib fwiw the `Cargo.lock`s aren't checked in for the example, but they aren't in `.gitignore` either it seems.

rib: right, probably because the examples were copied from rust-android-examples where I intentionally have Cargo.lock files.

should add to .gitignore

MarijnS95: Can drive-by fix that perhaps... but the CI is already green now.

Issue #92, opened 2023-06-27
Use common `Keycode` type across backends


rib:
There's a notable portability hazard currently due to the `game-activity` and `native-activity` backends exposing different `Keycode` types.

For example https://github.com/rust-windowing/winit/commit/918430979f8219648daade44796c00893e42fdd8#diff-9169a22d6397a250be741006cd857b8a575f804c74cc07e3a4fb8f3341606d9b and https://github.com/rust-windowing/winit/commit/4a36741f9ccaa03ca0842362dd1ebc24fd2a1b11 both recently assumed they could use the `ndk::event::Keycode` type which inadvertently broke building Winit against the `game-activity` backend.

We should either use `ndk::event::Keycode` in both backends or move the current `game-activity` `Keycode` type so it can be shared.

Also relates to https://github.com/rust-mobile/android-activity/issues/45

MarijnS95:
Perhaps also relates to #48 when considering reusing `ndk` types that look/feel too similar to their `game-activity` counterpart.

We could even rewrite parts of the C code to _make them_ similar if desired.

PR #121, opened 2023-09-24
Re-export Axis from the ndk for native-activity

fornwall:
Currently `android_activity::input::Axis` is always the enum defined by the `android-activity` crate.

This means that when `native-activity` is used, the below code does not currently compile, since the `Pointer` re-exported from the ndk expects an `Axis` also re-exported from the ndk:

```rust
let e: android_activity::input::MotionEvent = ...;
let pointer = motion_event.pointer_at_index(0);
let value = pointer.axis_value(android_activity::input::Axis::X);
```

### DIFF
--- a/android-activity/src/game_activity/input.rs
+++ b/android-activity/src/game_activity/input.rs
@@ -15,10 +15,12 @@
 
 use std::convert::TryInto;
 
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+
 use crate::activity_impl::ffi::{GameActivityKeyEvent, GameActivityMotionEvent};
 use crate::input::{
-    Axis, ButtonState, Class, EdgeFlags, KeyAction, KeyEventFlags, Keycode, MetaState,
-    MotionAction, MotionEventFlags, Source, ToolType,
+    ButtonState, Class, EdgeFlags, KeyAction, KeyEventFlags, Keycode, MetaState, MotionAction,
+    MotionEventFlags, Source, ToolType,
 };
 
 // Note: try to keep this wrapper API compatible with the AInputEvent API if possible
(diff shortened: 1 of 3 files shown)


rib:
Ah, yep, awkward.

This is a bit fiddly because the input API specifically differs quite significantly between the `game-activity` and `native-activity` backends and we have repeatedly hit issues from trying to expose compatible APIs using different types.

For implementation convenience the `native-activity` backend used to more-directly expose all the `ndk` input types but this has repeatedly led to compatibility issues between the `native-activity` and `game-activity` backends where the `game-activity` types haven't matched the `ndk` API exactly or where downstream crates have just assumed they can use the `ndk` types and then not been compatible with the `game-activity` backend at all.

Ref: https://github.com/rust-mobile/android-activity/pull/107 and https://github.com/rust-mobile/android-activity/issues/92

Because of this I've generally tried to switch away from using any of the ndk crate types here unless they can also be used with the game-activity backend too and so I'm thinking we should probably instead double-down with consolidating input types and look at wrapping the `Pointer` type so we can hide the `ndk` `Pointer` as an implementation detail for the `native-activity` backend.

`Pointer` and `PointersIter` are essentially the last two `ndk` input types that are still exposed publicly.

In a similar way, `Pointer::tool_type()` still poses a portability hazard between the `native-activity` and `game-activity` backends.

fornwall:
Thanks for sharing the info!

> we should probably instead double-down with consolidating input types and look at wrapping the Pointer type so we can hide the ndk Pointer as an implementation detail for the native-activity backend.

Makes sense! Created https://github.com/rust-mobile/android-activity/pull/122 as an initial iteration on that, so closing this PR in favour of that one!

PR #124, opened 2023-09-26
Enable CI on all branches

fornwall:
Enabling CI on all branches makes it smoother for external contributors, who can check CI status before creating a PR and awaiting approval of a workflow run.

For pull requests the behaviour in this repository is the same - a workflow run on a PR still requires approval. But for forks, this helps out with faster CI feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: ci
 
 on:
   push:
-    branches: [main]
+    branches: '*'
   pull_request:
 
 env:


rib:
Thanks, yeah the other thing is that the default is that actions have to be approved for 'first time contributors'. 

I've also changed that to approve to 'first time contributors who are new to github'

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: ci
 
 on:
   push:
-    branches: [main]
+    branches: '*'
```
MarijnS95, line 5:
```suggestion
```

MarijnS95 (changes requested):

PR #128, opened 2023-10-14
Upgrade to stable `ndk 0.8` and `ndk-sys 0.5` releases

MarijnS95:


rib (approved):
Thanks this looks good to me

PR #130, opened 2023-10-15
Upgrade to ndk-sys 0.5.0 and ndk 0.8.0

rib:


MarijnS95:
Duplicate of #128?

MarijnS95 (changes requested):
This won't compile due to breaking changes, and #128 is already open to correct that.

rib:
ah yep, np will merge that one thanks

rib:
Closing as dup of #128

PR #135, opened 2023-10-16
Set `workspace.resolver = "2"` and avoid cargo warning

rib:
Since we're using `edition = "2021"` in `android-activity/Cargo.toml` (which defaults to the v2 resolver) then cargo warns that this isn't consistent with the v1 resolver default for the top-level workspace.


MarijnS95:
Nice, yeah this warning started to show for a bunch of crates for me since Rust 1.70 or something close to that.

PR #138, opened 2023-10-16
Expose MotionEvent::action_button() state

rib:
This exposes the button associated with a button press or release action.

The PR also includes a separate copy&paste fix for `MotionEvent::action()` that was spotted while adding `action_button()`.


rib:
Cc: @MarijnS95

Ideally would have waited for review from you but also want to get this 0.5 release done and this doesn't seem controversial.

MarijnS95:
Sure thing, especially when things happen after midnight :zzz:.

PR #143, opened 2023-10-28
README: Update crate version in `Cargo.toml` example

MarijnS95:


rib (approved):
Thanks for these nice tidy ups!

Apart from needing to fixup the `NonNull::new()` change, this looks good to me.

PR #145, opened 2023-11-18
Name spawned threads

fornwall:
Name spawned threads to make things more clear during debugging and profiling.

[Max thread name is 15](https://stackoverflow.com/questions/5026531/thread-name-longer-than-15-chars), so we can't fit more descriptive names like `android-activity-main` (but open for some other naming scheme).

### DIFF
--- a/android-activity/src/game_activity/mod.rs
+++ b/android-activity/src/game_activity/mod.rs
@@ -926,7 +926,7 @@ pub unsafe extern "C" fn _rust_glue_entry(native_app: *mut ffi::android_app) {
         };
 
         thread::Builder::new()
-            .name("android-stdio".to_string())
+            .name("stdio-to-logcat".to_string())
             .spawn(move || {
                 let tag = CStr::from_bytes_with_nul(b"RustStdoutStderr\0").unwrap();
                 let mut reader = BufReader::new(file);
@@ -960,7 +960,7 @@ pub unsafe extern "C" fn _rust_glue_entry(native_app: *mut ffi::android_app) {
         unsafe {
             // Name thread - this needs to happen here after attaching to a JVM thread,
             // since that changes the thread name to something like "Thread-2".
-            let thread_name = CStr::from_bytes_with_nul(b"android-main\0").unwrap();
+            let thread_name = CStr::from_bytes_with_nul(b"android_main\0").unwrap();
             libc::pthread_setname_np(libc::pthread_self(), thread_name.as_ptr());
 
             let app = AndroidApp::from_ptr(NonNull::new(native_app).unwrap(), jvm.clone());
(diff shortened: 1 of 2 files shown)


MarijnS95:
This probably conflicts with https://github.com/rust-mobile/android-activity/pull/133, which IMO should be merged first to stay chronological.

MarijnS95 (approved):

fornwall:
> Since we know we're running on Android maybe we can drop the 'android-' prefix even, unless the Java main thread is already called 'main' perhaps?

The reasoning for the `android-` prefix was to improve clarity for someone listing thread names for an application, just having `main` and `stdio` is a bit more nondescriptive/generic.

But how about:
- `native-activity` and `game-activity` for the main threads - in that way one call also clearly tell if the game or native activity is used.
- `stdio-to-logcat` as the stdio-forwarder thread - shows what it is doing.
?

rib:
Yeah, `stdio-to-logcat` sounds a bit more descriptive of that thread's function.

I think `main` would be a pretty typical name for an application's main thread (and what I'd also find intuitive if I was looking at a list of threads in a profiler) but my main :) concern with that on Android is that that it's common to refer to the application's Java thread as the 'main' thread and maybe that's even what it's named at the system level?

I wouldn't be super keen on having 'activity' in the name since the Activity is really something that exists on the JVM side of things which might then be misleading.

Something like `android_main` might be appropriate if we need to disambiguate from a Java 'main' thread where this is the name of the actual `fn android_main()` native function that's called.

fornwall:
> I wouldn't be super keen on having 'activity' in the name since the Activity is really something that exists on the JVM side of things which might then be misleading.

Agreed!

> Something like android_main might be appropriate if we need to disambiguate from a Java 'main' thread where this is the name of the actual fn android_main() native function that's called.

Sounds good - let's use `android_main`?

Pushed an updated to use `android_main` and `stdio-to-logcat` as thread names now.

rib:
Okey, @MarijnS95's updated PR for #133 ended up naming the stdio-to-logcat thread as proposed here (it made sense to do the rename while anyway factoring out a shared utility for spawning that thread), so its just the `android_main` thread left to name here. Yeah I think 'android_main' seems reasonable to me based on the name of the top-level native function for that thread.

fornwall:
> Okey, @MarijnS95's updated PR for https://github.com/rust-mobile/android-activity/pull/133 ended up naming the stdio-to-logcat thread as proposed here (it made sense to do the rename while anyway factoring out a shared utility for spawning that thread), so its just the android_main thread left to name here

👍 - merge conflict has now been resolved.

MarijnS95 (approved):

rib (approved):
Thanks, this looks good to me

PR #147, opened 2023-12-13
native-activity/input: OR with `EVENT_ACTION_MASK` when extracting action

### LINKED ISSUE #146: Can't get multitouch to work with native activity
Not sure who to blame here (me, android-activity or winit), but after switching from `ndk_glue` to this crate + latest winit version my application no longer receives `Touch` start/end/cancel events for the second finger, only move ones; and even those are broken (position is in the center of two fingers)

ArthurCose:
Partially fixes #146. Possibly fully fixes the issue as I haven't encountered any issues with multitouch positions with winit on Android.

### DIFF
--- a/android-activity/src/native_activity/input.rs
+++ b/android-activity/src/native_activity/input.rs
@@ -55,7 +55,8 @@ impl<'a> MotionEvent<'a> {
         // `MotionAction` enum that we share between backends, which may also
         // capture unknown variants added in new versions of Android.
         let action =
-            unsafe { ndk_sys::AMotionEvent_getAction(self.ndk_event.ptr().as_ptr()) as u32 };
+            unsafe { ndk_sys::AMotionEvent_getAction(self.ndk_event.ptr().as_ptr()) as u32 }
+                & ndk_sys::AMOTION_EVENT_ACTION_MASK;
         action.into()
     }
 


MarijnS95 (approved):
Agreed, this was wrongly converted from the NDK.

@rib regarding:

https://github.com/rust-mobile/android-activity/blob/672360c5e6ac55f2c9707885cd8938d23de79bfa/android-activity/src/native_activity/input.rs#L362-L364

I desperately want to fix this in the NDK as there are new values in r26 that I cannot add without an API break. A few things:

- I want to make these `non_exhaustive` to allow adding new values without API breaks;
- The NDK wrongly converts not-yet-handled values to `Unknown` which always has a special `_UNKNOWN` variant (with value `0`). I.e. the conversion is lossy;
- I don't really want to have an `Unknown(i32)` case to provoke updating the NDK instead. I don't know how common it is (for e.g. custom Android distros) to have custom/unknown values;
 (One big issue I face now is that there are a bunch of new - and super useful! - functions in Android 14 that are not yet in a released NDK, so I might generate from the latest sysroot as was suggested to me before);
- But that requires having a `Result` with `TryFromPrimitiveError`;
- Hence I do like that you've used the combination of `non_exhaustive` with an `Unknown` case that is `#[doc(hidden)]`.

After that is done, how do you feel about using the `ndk` types again, assuming `NativeActivity` and `GameActivity` expose the same values? If not, are there any known mismatching / missing values?

rib (approved):
Sorry for the delay with reviewing this.

This looks good to me - thanks!

rib:
> Pardon me, just passing by, but... why is `OR` in the PR title, when the added operator is `&`? 🤔

hmm, yeah title should have been updated and then I also copy & pasted that for the changelog, oops.

MarijnS95:
Yeah so the history in the repo will be wrong, but the PR title and ~changelog~ GitHub release are fixed to mention `AND` at least :)

rib:
> @rib side-note: the changelog uses `[0.2]` link labels, but those tags are never specified to point to a link, e.g. `[0.2]: https://github.com/rust-mobile/android-activity/releases/tag/v0.2.0`.

_yeah_ I guess we could turn those into links like that. The [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) reference that's linked in the changelog does that, so would be nice to be consistent with that.

> Yeah so the commit title in the repo and changelog will be wrong, but the PR title and GitHub release are fixed to mention AND at least :)

we can also fly-by fix up the changelog the next time that gets touched.

MarijnS95:
Exactly that, yes :)

PR #152, opened 2024-01-30
Specify `cargo-ndk` version in CI

fornwall:


fornwall:
Replaced by a better approach to specify `cargo +stable` when installing instead.

PR #155, opened 2024-04-26
Upgrade to `ndk-sys 0.6.0` and `ndk 0.9.0`

MarijnS95:
The next breaking `ndk` release puts a lot of emphasis in improving `enum`s to finally be marked `non_exhaustive`, and carry possible future values in `__Unknown(i32)` variants.  This removes the lossy conversions that previously required `android-activity` to redefine its types, which could all be removed again.

The `repr()` types have also been updated, as `enum` constants in C are translated to `u32` by default in `bindgen` even though they're commonly passed as `int` to every API function that consumes them.

### DIFF
--- a/android-activity/src/native_activity/input.rs
+++ b/android-activity/src/native_activity/input.rs
@@ -262,13 +262,8 @@ impl<'a> PointerImpl<'a> {
     #[inline]
     pub fn axis_value(&self, axis: Axis) -> f32 {
         let value: u32 = axis.into();
-        if let Ok(ndk_axis) = value.try_into() {
-            self.ndk_pointer.axis_value(ndk_axis)
-        } else {
-            // FIXME: We should also be able to query `Axis::__Unknown(u32)` values
-            // that can't currently be queried via the `ndk` `Pointer` API
-            0.0f32
-        }
+        let value = value as i32;
+        self.ndk_pointer.axis_value(value.into())
     }
 
     #[inline]
@@ -283,7 +278,8 @@ impl<'a> PointerImpl<'a> {
 
     #[inline]
     pub fn tool_type(&self) -> ToolType {
-        let value: u32 = self.ndk_pointer.tool_type().into();
+        let value: i32 = self.ndk_pointer.tool_type().into();
+        let value = value as u32;
         value.into()
     }
 }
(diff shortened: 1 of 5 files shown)


android-activity/src/native_activity/mod.rs
```diff
@@ -507,6 +507,7 @@ impl<'a> InputIteratorInner<'a> {
                     ndk::event::InputEvent::KeyEvent(e) => {
                         input::InputEvent::KeyEvent(input::KeyEvent::new(e))
                     }
+                    _ => todo!("NDK added a new type"),
```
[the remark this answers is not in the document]
MarijnS95: Yeah fair enough, we're currently only mapping `Key` and `Motion` here, and disregarding `Focus`/`Capture`/`Drag`/`TouchMode`. Those don't have any custom functions, I'm not sure if they're used at all or only with the very limited generic `AInputEvent_` functions that provide next to no info?

android-activity/src/native_activity/input.rs
```diff
@@ -262,13 +262,8 @@ impl<'a> PointerImpl<'a> {
...
         let value: u32 = axis.into();
-        if let Ok(ndk_axis) = value.try_into() {
-            self.ndk_pointer.axis_value(ndk_axis)
-        } else {
-            // FIXME: We should also be able to query `Axis::__Unknown(u32)` values
-            // that can't currently be queried via the `ndk` `Pointer` API
-            0.0f32
-        }
+        let value = value as i32;
+        self.ndk_pointer.axis_value(value.into())
```
MarijnS95, line 266: By the way, I wondered if maybe the repr type should have been fixed as `game-activity` is converting them as well. That matches what I did in the NDK to match the enum type to the function(s) that consume/generate it, not to the default `u32` that happens to be assigned to untyped `enum`s.

At some point that'll be fixed upstream anyway as they're retroactively fixing things to be `enum foo : int32_t` now that the NDK is always compiled with `clang`.

MarijnS95: But if I read things like this in `android-activity` again:

```rust
        // XXX: we use `AInputEvent_getSource` directly (instead of calling
        // ndk_event.source()) since we have our own `Source` enum that we
        // share between backends, which may also capture unknown variants
        // added in new versions of Android.
```

It feels like most if not all enums that are redefined are _not_ to capture differences with `game-activity`, but merely because the `ndk` crate was lagging behind Android (what the comment literally says).

Instead of tediously spending a lot of time doing the same "type rectification" here, I'd rather extend the `ndk` with all the missing variants (which we're now allowed to do non-breaking thanks to using the exact same `#[non_exhaustive] enum` setup as `android-activity`), compare with `android-activity` to make sure we're not missing out, and finally reduce some NIH :)

PR #156, opened 2024-04-26
Bump MSRV to 1.69.0 considering we can't build cargo ndk with 1.68

rib:
cargo ndk will fail to build with 1.68 due to a toml_edit dep.

Technically android-activity itself should still build with 1.68 but it's simpler to synchronize the `rust-version` with the minimum version that we actually test in CI (where we need to build cargo ndk)


MarijnS95:
`winit` is on `1.70` so this is all good.

MarijnS95 (approved):

Issue #153, opened 2024-02-14
panic on nightly: `from_raw_parts` safety contract violation


skibon02:
On latest nightly build my application fall into panic even before main function starts.
After figuring out, the problem was in file `glue.rs`:
https://github.com/rust-mobile/android-activity/blob/c9faa9c44e04152863cc0ea9b6982065e0346ba6/android-activity/src/native_activity/glue.rs#L341

here, the saved_state_in is `null` in my case, which makes this operation violating safety contract.

To reproduce, create any minimal application with empty `android_main` function with `native-activity` feature.

Potentially, game activity code also have this issue, but wasn't tested

PR #157, opened 2024-04-26
native-activity: Check for null saved_state_in pointer

rib:
Avoids calling `std::slice::from_raw_parts` with a null `saved_state_in` pointer.

Fixes: #153


rib:
I opened up a new PR since I wasn't able to push to the branch for https://github.com/rust-mobile/android-activity/pull/154

The commit should still attribute @skibon02 for the fix (thanks!)

PR #163, opened 2024-07-04
input: Replace open-coded types with `ndk::event` definitions

MarijnS95:
As documented in `// XXX` comments, and discussed before, the copy-pasted `input` types exist because the `ndk` crate "wasn't set up" to allow changes to `enum`s in a non-semver-breaking way (in particular, they were _not_ marked `#[non_exhaustive]`).  This issue was not reported upstream, and it took me to use `android-activity` before realizing so and fixing it in https://github.com/rust-mobile/ndk/pull/459.  Also, after comparing the copy-paste here against the `ndk` types, no new variants were ever added, completely defeating the purpose (even though new types were added to the ustream headers and will be added to te `ndk` crate).

Hence, remove all the open-coded types again and reexport them from the `ndk` crate.  Upstream not only made all `enum`s `#[non_exhaustive]`, but also fixes the `repr` types to not use the `u32` default for untyped enum constants, but in most cases `i32` when all function signatures take a signed argument.


MarijnS95:
Will rebase this as soon as #164 is in!

android-activity/src/input.rs
```diff
@@ -1,243 +1,14 @@
...
-    pub fn is_position_class(self) -> bool {
-        let class = SourceFlags::from_bits_truncate(self.into());
-        class.contains(SourceFlags::POSITION)
-    }
-    #[inline]
-    pub fn is_joystick_class(self) -> bool {
-        let class = SourceFlags::from_bits_truncate(self.into());
-        class.contains(SourceFlags::JOYSTICK)
-    }
-}
```
rib, line 119: This seems like a notable API change that we should document in the CHANGELOG. With the `ndk` API the class is exposed via `Source::class()` instead.

android-activity/src/input.rs
```diff
@@ -1,243 +1,14 @@
...
-// ndk_sys doesn't currently have the `TRACKBALL` flag so we define our
-// own internal class constants for now
-bitflags! {
-    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-    struct SourceFlags: u32 {
-        const CLASS_MASK = 0x000000ff;
-
-        const BUTTON = 0x00000001;
-        const POINTER = 0x00000002;
-        const TRACKBALL = 0x00000004;
```
rib, line 86: Compared to the `ndk` API, this notably has a different name. The Android SDK names this 'Trackball' whereas the `ndk` C API calls this `CLASS_NAVIGATION`. I happened to go with the SDK naming, so it could be worth noting this change in the CHANGELOG

rib (changes requested):
Thanks for looking at this, I think in general this seems reasonable to do at this point.

Originally, a notable part of the purpose for these types was just to provide types that could be shared between the `GameActivity` and `NativeActivity` backend. 

There have repeatedly been issues with directly exposing `ndk` input types in the android-activity API due to differences between `GameActivity` and `NativeActivity` and so I've generally been cautious about going back to exposing them. E.g. for a while the Winit Android backend wouldn't work with the game-activity backend because of stuff like this: https://github.com/rust-windowing/winit/blob/918430979f8219648daade44796c00893e42fdd8/src/platform_impl/android/mod.rs#L404

Since the input API is so central to the `android-activity` crate then in the past it's also been convenient to not be too tightly coupled to the `ndk` crate here, so I could make input API changes without being blocked on there being a new `ndk` release.

Maybe if there are some missing variants it's a good time to look at using shared types from the `ndk` crate if they can now be shared across both backends.

I think the main thing that seems to be missing atm are some CHANGELOG notes about the breaking changes. For completeness that should probably include a note about the `repr` changes

MarijnS95:
> There have repeatedly been issues with directly exposing `ndk` input types in the android-activity API due to differences between `GameActivity` and `NativeActivity`

We discussed this before, and as highlighted in this PR that those compatibility issues weren't brought up "upstream" at the `ndk` crate before, nor really seem to exist when I put the effort in comparing these enums to replace them. If anything the `ndk` already has more variants for some enums, and has a change scheduled to add anything that's still missing.

> E.g. for a while the Winit Android backend wouldn't work with the game-activity backend because of stuff like this: https://github.com/rust-windowing/winit/blob/918430979f8219648daade44796c00893e42fdd8/src/platform_impl/android/mod.rs#L404

We've since fixed that, which is specifically why this PR is open _now_ so I don't think it's fair to dredge that back up.

> Since the input API is so central to the `android-activity` crate then in the past it's also been convenient to not be too tightly coupled to the `ndk` crate here, so I could make input API changes without being blocked on there being a new `ndk` release.

Again, there are little to no differences at this time. I'm keeping the `struct`s on your side and only replacing the `enum`s and `bitflag`s anyway.

> Maybe if there are some missing variants it's a good time to look at using shared types from the `ndk` crate if they can now be shared across both backends.

That's exactly what this PR did :)

> I think the main thing that seems to be missing atm are some CHANGELOG notes about the breaking changes. For completeness that should probably include a note about the `repr` changes

Fixed.

android-activity/CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- input: Replaced custom types with their `ndk` crate equivalent.
+  > [!NOTE]
+  > These types existed because the `ndk` crate didn't provide them in an extensible way.  Now that they have the `#[non_exhaustive]` flag and contain a `__Unknown(T)` variant to result in lossless conversions, and not to mention use an ABI type that how its used by most functions (when the original constants were defined in a "typeless" way), the `ndk` types are used and reexported once again.
```
MarijnS95, line 11:
```suggestion
  > These types existed because the `ndk` crate didn't provide them in an extensible way.  Now that they have the `#[non_exhaustive]` flag and contain a `__Unknown(T)` variant to result in lossless conversions, and not to mention use an ABI type that  matches how it is being used by most functions (when the original constants were defined in a "typeless" way), the `ndk` types are used and reexported once again.
```

android-activity/CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+  > [!IMPORTANT]
+  > **Relevant breaking changes**:
+  > - `repr()` types for some `enum`s have changed to match the type that returned and consumed by functions interacting with this wrapper type.
```
MarijnS95, line 15:
```suggestion
  > - `repr()` types for some `enum`s have changed to match the type that is returned from and consumed by functions interacting with this wrapper type.
```

MarijnS95: I don't like the word `interacting` here but haven't found a concise way to say "a function that returns or uses one of these wrapper types".

MarijnS95: "referencing this wrapper type" perhaps?

MarijnS95: This seems best to me:

> - `repr()` types for some `enum`s have changed to match the ABI type that is used by most functions that are returning or consuming this wrapper type.

rib:
I'm having nagging second thoughts that we shouldn't have merged this as a breaking change :/

This change means we can't make a release of android-activity that will be compatible with winit 0.30.x

Knowing how long Winit's major release cycles can be that doesn't seem ideal

PR #164, opened 2024-07-29
Fix various codebase rots (stale CI, new Rust lints, broken MSRV checks by transitive dependency upgrades)

MarijnS95:
[Rust 1.80 from July 25th 2024] points out that `armv7` is not a known, valid value for the `target_arch` cfg variable.  This is confirmed by the docs not listing it either: https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch.

Hence drop this entirely, and rely purely on `target_arch = "arm"`.

Furthermore it also points out that we don't have a `test` and `api-level-30` feature: remove the unused `test` feature and reexpose `ConfigurationRef::screen_round()` which was behind a previously unsettable `feature = "api-level-30"`.

[Rust 1.80 from July 25th 2024]: https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html

---

EDIT: PR updated to fix **everything** that was holding back our CI.

### DIFF
--- a/android-activity/src/config.rs
+++ b/android-activity/src/config.rs
@@ -6,13 +6,14 @@ use ndk::configuration::{
     ScreenSize, Touchscreen, UiModeNight, UiModeType,
 };
 
-/// A (cheaply clonable) reference to this application's [`ndk::configuration::Configuration`]
+/// A runtime-replacable reference to [`ndk::configuration::Configuration`].
 ///
-/// This provides a thread-safe way to access the latest configuration state for
-/// an application without deeply copying the large [`ndk::configuration::Configuration`] struct.
+/// # Warning
 ///
-/// If the application is notified of configuration changes then those changes
-/// will become visible via pre-existing configuration references.
+/// The value held by this reference **will change** with every [`super::MainEvent::ConfigChanged`]
+/// event that is raised.  You should **not** [`Clone`] this type to compare it against a
+/// "new" [`super::AndroidApp::config()`] when that event is raised, since both point to the same
+/// internal [`ndk::configuration::Configuration`] and will be identical.
 #[derive(Clone)]
 pub struct ConfigurationRef {
     config: Arc<RwLock<Configuration>>,
--- a/android-activity/src/config.rs
+++ b/android-activity/src/config.rs
@@ -28,8 +28,6 @@ impl PartialEq for ConfigurationRef {
     }
 }
 impl Eq for ConfigurationRef {}
-unsafe impl Send for ConfigurationRef {}
-unsafe impl Sync for ConfigurationRef {}
 
 impl fmt::Debug for ConfigurationRef {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
--- a/android-activity/src/native_activity/glue.rs
+++ b/android-activity/src/native_activity/glue.rs
@@ -81,12 +81,14 @@ pub struct WaitableNativeActivityState {
     pub cond: Condvar,
 }
 
+// SAFETY: ndk::NativeActivity is also SendSync.
+unsafe impl Send for WaitableNativeActivityState {}
+unsafe impl Sync for WaitableNativeActivityState {}
+
 #[derive(Debug, Clone)]
 pub struct NativeActivityGlue {
     pub inner: Arc<WaitableNativeActivityState>,
 }
-unsafe impl Send for NativeActivityGlue {}
-unsafe impl Sync for NativeActivityGlue {}
 
 impl Deref for NativeActivityGlue {
     type Target = WaitableNativeActivityState;
@@ -223,7 +225,6 @@ pub struct NativeActivityState {
     /// Set as soon as the Java main thread notifies us of an
     /// `onDestroyed` callback.
     pub destroyed: bool,
-    pub redraw_needed: bool,
     pub pending_input_queue: *mut ndk_sys::AInputQueue,
     pub pending_window: Option<NativeWindow>,
 }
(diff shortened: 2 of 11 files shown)


android-activity/src/native_activity/glue.rs
```diff
@@ -75,18 +75,20 @@ pub enum State {
 
 #[derive(Debug)]
 pub struct WaitableNativeActivityState {
+    // TODO: Unnecessary NIHism: why not use ndk::NativeActivity?
```
rib, line 78:
```suggestion
    // TODO: use ndk::NativeActivity?
```

I half guess there might have been a reason at the time, but don't really recall - it sounds like it would be fine to change.

rib (approved):
Thanks, this all looks good to me!

android-activity/src/native_activity/glue.rs
MarijnS95: I half-guessed what was going on in the mean-time: the NDK crate that I inherited still has rough/incorrect edges in places I haven't been looking: https://github.com/rust-mobile/ndk/issues/478

And since afaik you ported this glue code straight from C, it was easier to append `unsafe { ffi::<original code> }` than rewriting it with `ndk` primitives.

android-activity/src/config.rs
```diff
@@ -28,8 +28,6 @@ impl PartialEq for ConfigurationRef {
     }
 }
 impl Eq for ConfigurationRef {}
-unsafe impl Send for ConfigurationRef {}
-unsafe impl Sync for ConfigurationRef {}
```
MarijnS95, line 32: @rib looking at this code again, any clue why it exists at all? `ndk::Configuration` is already marked `Send + Sync` because all functions/operations on it only read underlying data without modifying the pointer, except for `.copy()` which is specifically marked as `&mut`. We don't implement the setters yet but that's something I plan to do.

Even if you want to keep this wrapper struct, it could/should be a simple `Arc<Configuration>` newtype _without_ lock, and a simple `Deref` to the underyling `&Configuration`. That would get us rid of having to forward the `api-level-30` feature too.

MarijnS95: Looking more closely, `android-activity` doesn't have it because of the first two paragraphs in its documentation, but only because of the third: to have interior mutability when an app raises that the configuration has changed.

This seems like a very scary thing to me, but I'm not going to look into that as part of this PR.

MarijnS95:
@rib how do we want to deal with this for our MSRV:

```
error: package `num_enum v0.7.3` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.69.0
Either upgrade to rustc 1.70.0 or newer, or use
cargo update -p num_enum@0.7.3 --precise ver
where `ver` is the latest version of `num_enum` supporting rustc 1.69.0
```

Shall we run a `cargo +nightly generate-lockfile -Zminimal-versions` before running the MSRV check?

MarijnS95:
@notgull would love to have your review on this as well, now that the CI is made green again.

notgull:
Please rerun the CI, I don't think this fixes the "invalid configuration" errors I was getting.

MarijnS95:
@notgull the CI has ran quite a few times on a personal branch before I force-pushed the final

Otherwise, please link your CI failure. There is only one run from your PR: https://github.com/rust-mobile/android-activity/actions/workflows/ci.yml?query=actor%3Anotgull

And that run failed because of specifically the errors in this PR: https://github.com/rust-mobile/android-activity/actions/runs/11771937579

notgull (approved):
I don't see anything objectionable.

android-activity/src/config.rs
MarijnS95: Hold up, I'm marking this as unresolved because I removed one important paragraph too many, that deserves a much bigger ~highlight~ warning flags. Folks are expecting a regular NDK `Configuration` here to compare what has changed since the last event to omit potential costly updates, yet never observe any change since the `app.config().clone()` that they previously stored is simply a wrapper around an `Arc` referring to the ""new"" config that they just pulled out of `app.config()` 🤦

android-activity/src/config.rs
```diff
@@ -6,13 +6,14 @@ use ndk::configuration::{
...
-/// This provides a thread-safe way to access the latest configuration state for
-/// an application without deeply copying the large [`ndk::configuration::Configuration`] struct.
+/// # Warning
 ///
-/// If the application is notified of configuration changes then those changes
-/// will become visible via pre-existing configuration references.
+/// The value held by this reference **will change** with every [`super::MainEvent::ConfigChanged`]
+/// event that is raised.  You should **not** [`Clone`] this type to compare it against a
+/// "new" [`super::AndroidApp::config()`] when that event is raised, since both point to the same
+/// internal [`ndk::configuration::Configuration`] and will be identical.
```
MarijnS95, line 16: @rib I don't have any alternative to propose to users to compare configurations besides calling the expensive `fn copy()` to make a deep-clone of `ndk::configuration::Configuration` which exactly defeats the purpose of initially making this "cheap to copy".

By having this an updatable `RwLock` rather than a simple `Arc` it is no longer possible to compare for configuration differences and prevent possibly expensive reconfiguration/updates in the users' loops. If this was "just an `Arc`" users could do exactly that while still "cloning" the configuration around cheaply.

In any case, every time the callback fires a new `AConfiguration` is allocated and filled anyway, rather than "reading into" the existing `AConfiguration`. But the docs at https://developer.android.com/ndk/reference/group/configuration#aconfiguration_fromassetmanager are incredibly vague by mentioning `Create and return a new AConfiguration` while it has to take an existing allocated (but possibly new/empty?) `AConfiguration`.

MarijnS95: Postponing to #180, to at least get this in and unblock future contributions.

PR #165, opened 2024-08-03
Add `package.include` to `Cargo.toml`

daxpedda:
See [the Cargo documentation on `package.include`](https://doc.rust-lang.org/1.80.0/cargo/reference/manifest.html#the-exclude-and-include-fields).

When using `cargo publish` every file is included and uploaded, even the shell files used to generate bindings not necessary during build time. Using `package.include` we can make sure that this doesn't happen and only what the user needs to compile the package or generate local documentation is included.


ardocrat:
Good catch, but I guess you can setup this at your application, not at lib level?

daxpedda:
> Good catch, but I guess you can setup this at your application, not at lib level?

Not sure what you mean, this is about crates.io releases.
Or do people use `cargo publish` for something else?

android-activity/Cargo.toml
```diff
@@ -9,6 +9,13 @@ repository = "https://github.com/rust-mobile/android-activity"
 documentation = "https://docs.rs/android-activity"
 description = "Glue for building Rust applications on Android with NativeActivity or GameActivity"
 license = "MIT OR Apache-2.0"
+include = [
+    "/build.rs",
+    "/docs",
```
MarijnS95, line 14: This folder only contains `LICENSE-APACHE` and `LICENSE-MIT`, we should move those to the crate root (and delete this folder) to replace `LICENSE`.

android-activity/Cargo.toml
```diff
@@ -9,6 +9,13 @@ repository = "https://github.com/rust-mobile/android-activity"
...
+    "/game-activity-csrc",
+    "/LICENSE",
+    "/src/**/*.rs",
```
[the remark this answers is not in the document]
daxpedda: Done.

android-activity/Cargo.toml
daxpedda: Done.

android-activity/Cargo.toml
```diff
@@ -9,6 +9,13 @@ repository = "https://github.com/rust-mobile/android-activity"
 documentation = "https://docs.rs/android-activity"
 description = "Glue for building Rust applications on Android with NativeActivity or GameActivity"
 license = "MIT OR Apache-2.0"
+include = [
+    "/build.rs",
+    "/docs",
+    "/game-activity-csrc",
+    "/LICENSE*",
```
MarijnS95, line 16: That GLOB in there is really hard to notice :)

Remind me, `*` is _zero_-or-more characters right?

MarijnS95 (approved):
One more question: how is `README.md` from the root of this repo getting into the `android-activity` subfolder I don't see a symlink of sorts but it's in there: https://docs.rs/crate/android-activity/latest/source/README.md

android-activity/Cargo.toml
daxpedda: Yes!

daxpedda: [Git says](https://git-scm.com/docs/gitignore):
> An asterisk "`*`" matches anything except a slash.

[Cargo says](https://doc.rust-lang.org/1.80.0/cargo/reference/manifest.html#the-exclude-and-include-fields):
> `*` matches zero or more characters except `/`.

daxpedda:
> One more question: how is `README.md` from the root of this repo getting into the `android-activity` subfolder? I don't see a symlink of sorts but it's in the packaged crate: https://docs.rs/crate/android-activity/latest/source/README.md

Its getting it from the `package.readme` field, which correctly points to the README file.
But its missing in [the documentation](https://doc.rust-lang.org/1.80.0/cargo/reference/manifest.html#the-exclude-and-include-fields) that the README is included by default, so that's interesting.

MarijnS95:
> Its getting it from the `package.readme` field, which correctly points to the README file.

Ah, I didn't see that there's a `../` in there which, typically, it's not allowed to go outside of the crate root. Guess `cargo` makes an exception for singular files with an explicit location.

> But its missing in [the documentation](https://doc.rust-lang.org/1.80.0/cargo/reference/manifest.html#the-exclude-and-include-fields) that the README is included by default, so that's interesting.

Yeah, that's correct.

rib (approved):
Nice, this looks good to me too, thanks!

rib:
Since merging the update to GameActivity 4.0, I've tweaked this to include `/android-games-sdk` instead of `/game-activity-csrc` (which no longer exists)

PR #168, opened 2024-09-16
fix: Correct misspelled `teriary()`. Deprecate it in favor of `tertiary()`.

Enyium:


MarijnS95:
Note that this issue is moved to the `ndk` crate (and fixed in https://github.com/rust-mobile/ndk/pull/486) if/when https://github.com/rust-mobile/android-activity/pull/163 is merged.

MarijnS95:
Fixes via https://github.com/rust-mobile/android-activity/pull/163 instead, but thanks for bringing this to our attention!

Issue #170, opened 2024-11-02
`ALooper_pollAll` is marked deprecated by Android NDK


wuwbobo2021:
I have found this discussion occasionally: <https://github.com/android/ndk/discussions/2020>.

Check <https://developer.android.google.cn/ndk/reference/group/looper#alooper_pollall>:

[Deprecated. ] Calls to `ALooper_pollAll` should be replaced with `ALooper_pollOnce`. If you call `ALooper_pollOnce` in a loop, you must treat all return values as if they also indicate `ALOOPER_POLL_WAKE`.

MarijnS95:
On the NDK side this is handled in https://github.com/rust-mobile/ndk/pull/491. If we didn't replace it in `android-activity` yet, a crate upgrade will cause a deprecation warning.

wuwbobo2021:
I had encountered a memory leak problem while debugging the initial version of `android-usbser-rs`. A loop polling for USB transfer data causes the problem, and it crashed on older Android versions (< 8.0).

I did an experiement and realized that `attach_current_thread_permanently()` prevents automatic de-allocation of JNI local references, unless making use of local frames or the `AutoLocal` wrapper. Is there a solution with better performance?

- <https://docs.rs/jni/latest/jni/struct.JNIEnv.html#references-and-lifetimes>
- <https://docs.rs/jni/latest/jni/struct.JNIEnv.html#method.auto_local>

Test program:

```
[package]
name = "android-simple-test"
version = "0.1.0"
edition = "2021"
publish = false

[dependencies]
jni = "0.21"
ndk-context = "0.1"
android-activity = { version = "0.6", features = ["native-activity"] }

[lib]
name = "android_simple_test"
crate-type = ["cdylib"]
path = "main.rs"

[package.metadata.android]
package = "com.example.android_simple_test"
build_targets = [ "aarch64-linux-android" ]

[package.metadata.android.sdk]
min_sdk_version = 16
target_sdk_version = 30
```

```
#[no_mangle]
fn android_main(_: android_activity::AndroidApp) {
    let mut i: i32 = 0;
    loop {
        let _ = with_jni_env_ctx(|env, _| {
            let _ = env.new_string(format!("test {}", i))
                // .map(|o| env.auto_local(o))
                .unwrap();
            Ok(())
        });
        i = i.wrapping_add(1);
        if i == 1000_000 {
            println!("Not crashed");
        }
    }
}

use jni::errors::Error;
[...]

PR #171, opened 2024-11-11
Add a way to easily convert an AndroidAppWaker into a Waker

notgull:
This commit adds a "waker()" method to AndroidAppWaker. It converts it
into an `std::task::Waker`, which is the type of waker used by
asynchronous tasks for scheduling. The goal is to allow AndroidAppWaker
to be easily used to set up an asynchronous context.

The implementation is very efficient, owing to the "static pointer"
style of coding already used. The "wake" function just calls
ALooper_wake, and cloning/dropping the waker is just a copy.

Discussion questions:

- Should `waker()` take `&self` or `self`? I chose the latter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target
+Cargo.lock
--- a/android-activity/src/game_activity/mod.rs
+++ b/android-activity/src/game_activity/mod.rs
@@ -8,6 +8,7 @@ use std::ptr;
 use std::ptr::NonNull;
 use std::sync::Weak;
 use std::sync::{Arc, Mutex, RwLock};
+use std::task::{RawWaker, RawWakerVTable, Waker};
 use std::time::Duration;
 
 use libc::c_void;
@@ -117,6 +118,27 @@ impl AndroidAppWaker {
         unsafe {
             ALooper_wake(self.looper.as_ptr());
         }
+    }
+
+    /// Creates a [`Waker`] that wakes up the [`AndroidApp`].
+    ///
+    /// This is useful for using this crate in `async` environments.
+    ///
+    /// [`Waker`]: std::task::Waker
+    pub fn waker(self) -> Waker {
+        const VTABLE: RawWakerVTable = RawWakerVTable::new(clone, wake, wake, drop);
+
+        unsafe fn clone(data: *const ()) -> RawWaker {
+            RawWaker::new(data, &VTABLE)
+        }
+
+        unsafe fn wake(data: *const ()) {
+            ndk_sys::ALooper_wake(data as *const _ as *mut _)
+        }
+
+        unsafe fn drop(_: *const ()) {}
+
+        unsafe { Waker::from_raw(RawWaker::new(self.looper.as_ptr() as *const (), &VTABLE)) }
     }
 }
 
--- a/android-activity/src/native_activity/mod.rs
+++ b/android-activity/src/native_activity/mod.rs
@@ -6,6 +6,7 @@ use std::panic::AssertUnwindSafe;
 use std::ptr;
 use std::ptr::NonNull;
 use std::sync::{Arc, Mutex, RwLock, Weak};
+use std::task::{RawWaker, RawWakerVTable, Waker};
 use std::time::Duration;
 
 use libc::c_void;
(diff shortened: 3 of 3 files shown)


notgull:
Build failure is due to codebase rot

MarijnS95:
https://github.com/rust-mobile/android-activity/pull/164#issuecomment-2327429638

Yeah, we're really lacking maintenance here, unfortunately. As always, I'm happy to submit PRs to tackle issues and implement features every once in a while in between other tasks, but **need a second maintainer for review and approval**.

@rib can you update us on your availability and commitment towards this crate, or should we seek for a third maintainer to fill this review-gap?

MarijnS95:
Regarding the PR itself, note that the result of a looper wake is very loosely defined and not to be relied on: #170

If there's any explicit `async` task spawning and progress monitoring that needs to happen on an `ALooper`, I'd prefer if we add an abstraction based on file descriptors or callbacks which, unlike waking, can be explicitly handled.

notgull:
Interesting. Does it have a chance of spuriously not waking up the event loop? `winit` currently assumes it always does.

MarijnS95:
@notgull according to their documentation, it would get batched up with other events, such that the event loop is "active" / "awake" but without ever seeing it return `WAKE` because it got "folded into" the other events.

`winit` currently relies on `Poll::Wake` to trigger a user-requested redraw:

https://github.com/rust-windowing/winit/blob/ae4c449670674d8ac0d6d8754caf3fe5f4954c25/src/platform_impl/android/mod.rs#L586-L602

But it also checked these before entering a `poll_events()`: https://github.com/rust-windowing/winit/blob/ae4c449670674d8ac0d6d8754caf3fe5f4954c25/src/platform_impl/android/mod.rs#L562-L565

With these findings, it'll miss events on `redraw_flag` and `proxy_wake_up`, unless we hoist those checks out of the `match` block for any `Poll`/`Err` returned out of `poll_events()`. Or if we want to be more specific, attach a pipe or eventfd to be guaranteed a consistent result whenever our channel into the `Looper` signaled anything.

notgull:
I would prefer to just use eventfd.

MarijnS95:
Same preference here, it's much more explicit than pretending every `Poll` could have a `Wake` folded into it.

rib:
Sorry for only just coming around to look at this.

This seems reasonable.

Re: #170 and the way that wake events are documented: the point is just that when `_wake()` is used to wake up a Looper then although it does guarantee a wake up (you don't have to worry about it not waking) it's not guaranteed that you _specifically_ get a `::Wake` event if there are other events that can be delivered after the wake up.

Regarding how winit handles `Wake` events, winit is doing the right thing already I believe.

Winit never _specifically_ requires a `::Wake` event for anything - it's kind of the opposite and it will _ignore_ `::Wake` events when it currently has nothing it needs to do.

In itself a `Wake` event is the the last-resort event that's delivered after the even loop wakes after calling `epoll_wait` if there is _nothing_ else to report (such as a timeout or an idle callback or some other file descriptor event).

So from Winit's point of view a Wake event on its own doesn't represent anything interesting that needs to be handled, unless it knows that it requested a wake up.

The android backend for Winit effectively tracks when it's waiting for a wake up and so this code in Winit:

```rust
                android_activity::PollEvent::Wake => {
                    // In the X11 backend it's noted that too many false-positive wake ups
                    // would cause the event loop to run continuously. They handle this by
                    // re-checking for pending events (assuming they cover all
                    // valid reasons for a wake up).
                    //
                    // For now, user_events and redraw_requests are the only reasons to expect
                    // a wake up here so we can ignore the wake up if there are no events/requests.
                    // We also ignore wake ups while suspended.
                    self.pending_redraw |= self.redraw_flag.get_and_reset();
                    if !self.running
                        || (!self.pending_redraw && !self.user_events_receiver.has_incoming())
                    {
                        return;
                    }
                }
```
is about ignoring Wake events when Winit has nothing it needs to do.

In other words, an `ALooper_wake()` may lead to any event (it's only promising that `epoll_wait` or `_pollOnce` will unblock and return with _an_ event in a finite amount of time. This is why the Android NDK docs clarify that you should treat any event like a `Wake` event.  _I suppose in the past they made the mistake of implying in their docs that `_wake()` will always result in a `Wake` which is not the case_.

So in Winit specifically, it only has that special case for ignoring some Wake events and then apart from that, all events (including Wake events that are not ignored) will result in the same `single_iteration` code running.

rib:
Hi @notgull, can you maybe give a bit of context for closing this PR? are you maybe no longer interested in the feature?

If I were to merge the updated branch above, are you OK with keeping your attribution or should I remove it?

PR #184, opened 2025-03-10
Build-test documentation and fix broken doc samples

### LINKED ISSUE #182: Example code for AndroidApp::activity_as_ptr does not compile
I'm trying to figure out how to get the native activity in my app code. For example, when I want to get access to my app's internal files directory. 

The [activity_as_ptr](https://docs.rs/android-activity/0.6.0/android_activity/struct.AndroidApp.html#method.activity_as_ptr) method looks promising, but that code causes an error:

```rust
let activity = unsafe { JObject::from_raw(app.activity_as_ptr()) };
```

```text
error[E0308]: mismatched types
  --> my-app/src/lib.rs:75:47
   |
75 |     let activity = unsafe { JObject::from_raw(app.activity_as_ptr()) };
   |                             ----------------- ^^^^^^^^^^^^^^^^^^^^^ expected `*mut _jobject`, found `*mut c_void`
   |                             |
   |                             arguments to this function are incorrect
   |
   = note: expected raw pointer `*mut _jobject`
              found raw pointer `*mut c_void`

```

MarijnS95:
Fixes #182

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,10 +96,15 @@ jobs:
         run: >
           cargo ndk -t arm64-v8a doc --no-deps
 
-      - name: Tests (host build-testing)
+      - name: Build doctests
+        # All doctests are set to no_run, because they require running in the
+        # context of an Android app.
+        # Only run on stable because cross-compiling doctests is only supported
+        # since Rust 1.89.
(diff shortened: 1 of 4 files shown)


rib (approved):
Looks good to me, thanks

Nice to know that doc tests can be cross compiled now.

PR #185, opened 2025-03-11
Fix GameActivity touch inputs in the presence of a stylus

jb55:
I noticed touch inputs weren't working on my Galaxy S9 tab device which
includes a stylus. The value was 0xD002 for source, which seems to be:

```
SOURCE_CLASS_POINTER     0b0000 0000 0000 0010
SOURCE_STYLUS            0b0100 0000 0000 0000
SOURCE_BLUETOOTH_STYLUS  0b1000 0000 0000 0000
SOURCE_TOUCHSCREEN       0b0001 0000 0000 0000
----------------------------------------------
                0xD002 = 0b1101 0000 0000 0010
```

Let's modify default_motion_filter to use bitwise AND (&) instead of
equality (==) when checking SOURCE_TOUCHSCREEN. This ensures stylus
inputs (which combine touchscreen and stylus flags) are not discarded,
while still filtering non-touch events.

### DIFF
--- a/android-activity/game-activity-csrc/game-activity/native_app_glue/android_native_app_glue.c
+++ b/android-activity/game-activity-csrc/game-activity/native_app_glue/android_native_app_glue.c
@@ -29,11 +29,11 @@
 #define NATIVE_APP_GLUE_KEY_EVENTS_DEFAULT_BUF_SIZE 4
 
 #define LOGI(...) \
-    ((void)__android_log_print(ANDROID_LOG_INFO, "threaded_app", __VA_ARGS__))
+    ((void)__android_log_print(ANDROID_LOG_INFO, "android_activity", __VA_ARGS__))
 #define LOGE(...) \
-    ((void)__android_log_print(ANDROID_LOG_ERROR, "threaded_app", __VA_ARGS__))
+    ((void)__android_log_print(ANDROID_LOG_ERROR, "android_activity", __VA_ARGS__))
 #define LOGW(...) \
-    ((void)__android_log_print(ANDROID_LOG_WARN, "threaded_app", __VA_ARGS__))
+    ((void)__android_log_print(ANDROID_LOG_WARN, "android_activity", __VA_ARGS__))
 #define LOGW_ONCE(...)                                        \
     do {                                                       \
         static bool alogw_once##__FILE__##__LINE__##__ = true; \
@@ -46,7 +46,7 @@
 /* For debug builds, always enable the debug traces in this library */
 #ifndef NDEBUG
 #define LOGV(...)                                                   \
-    ((void)__android_log_print(ANDROID_LOG_VERBOSE, "threaded_app", \
+    ((void)__android_log_print(ANDROID_LOG_VERBOSE, "android_activity", \
                                __VA_ARGS__))
 #else
 #define LOGV(...) ((void)0)
--- a/android-activity/game-activity-csrc/game-activity/native_app_glue/android_native_app_glue.c
+++ b/android-activity/game-activity-csrc/game-activity/native_app_glue/android_native_app_glue.c
@@ -264,7 +264,7 @@ static bool default_key_filter(const GameActivityKeyEvent* event) {
 
 static bool default_motion_filter(const GameActivityMotionEvent* event) {
     // Ignore any non-touch events.
-    return event->source == SOURCE_TOUCHSCREEN;
+    return event->source & SOURCE_TOUCHSCREEN;
 }
 
 // --------------------------------------------------------------------


MarijnS95:
could we perhaps just take a more recent source of the [upstream GameSDK](https://android.googlesource.com/platform/frameworks/opt/gamesdk.git/), which fixed this a long time ago here?

https://android.googlesource.com/platform/frameworks/opt/gamesdk.git/+/d32d41d2cdf7f7f62b0a906e5614676061099583%5E%21/#F0

MarijnS95:
@rib would know best how to continue migrating and updating this crate, I only use (and feel comfortable reviewing/approving) native-activity parts.

jb55:
> could we perhaps just take a more recent source of the [upstream GameSDK](https://android.googlesource.com/platform/frameworks/opt/gamesdk.git/), which fixed this a long time ago here?
> 
> https://android.googlesource.com/platform/frameworks/opt/gamesdk.git/+/d32d41d2cdf7f7f62b0a906e5614676061099583%5E%21/#F0

Oh i had no idea this was sourced elsewhere

jb55:
Could this be why i’m seeing buggy stuff here as well?

- https://github.com/rust-mobile/android-activity/issues/186

Not sure where to go from here, but if its a matter of updating something from upstream at least i could try that

MarijnS95:
We've had quite a few reports that the vendored `GameActivity` code is buggy and that we ["should have done our due-diligence to report everything upstream"](https://github.com/rust-mobile/android-activity/pull/94#issuecomment-2622967812), so yeah it's quite likely that there are lot of bugs in there.

I'm still using `NativeActivity`, so no clue about everything that's going on with `GameActivity`.

jb55:
> We've had quite a few reports that the vendored `GameActivity` code is buggy and that we ["should have done our due-diligence to report everything upstream"](https://github.com/rust-mobile/android-activity/pull/94#issuecomment-2622967812), so yeah it's quite likely that there are lot of bugs in there.
> 
> I'm still using `NativeActivity`, so no clue about everything that's going on with `GameActivity`.

From what i was told, NativeActivity is deprecated and it was difficult to get the keyboard to work. I am able to show the keyboard and process text events with GameActivity through winit to egui. If i can resolve these remaining issues i might be able to actually launch our egui app on android, and hopefully open up egui dev on mobile a bit.

https://github.com/damus-io/notedeck

rib:
> We've had quite a few reports that the vendored `GameActivity` code is buggy and that we ["should have done our due-diligence to report everything upstream"](https://github.com/rust-mobile/android-activity/pull/94#issuecomment-2622967812), so yeah it's quite likely that there are lot of bugs in there.

fwiw, I intentionally didn't reply to that comment while it came across as kinda rude / accusatory. I have opened various upstream issues with GameActivity in the past.

`android-activity` doesn't really support you using arbitrary versions of `GameActivity` since the Java code is conceptually tightly coupled with the native part and if you're using `android-activity` then the native part is fixed because it's vendored in-tree as part of the android-activity build.

The comment about using `GameActivity` via a prefab is suspicious because it wouldn't really make sense imho to try and use the prefab magic to link the upstream native code for GameActivity since we have to patch that code to integrate with `android-activity` and you would also presumably end up linking in the native GameActivity code into your application twice. (i.e. the prefab would link in a copy of the code under `android-activity/game-activity-csrc` https://github.com/rust-mobile/android-activity/tree/main/android-activity/game-activity-csrc)

Currently if you're using the GameActivity backend you should ideally use version 2.0.2 (ref: https://github.com/rust-mobile/android-activity/pull/88).

If you use another, newer, version it's possible it _might_ work but that would depend on whether they have changed the internal ABI between their native code and JNI which isn't something they would be required to maintain.

jb55:
Looks like latest alpha version is 4.2.0?

https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/044fd03c4a7d3b75aeb6ca2bd7fb6155d2cdb787

https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/044fd03c4a7d3b75aeb6ca2bd7fb6155d2cdb787%5E%21/#F0

not sure what I should target. latest and greatest?

rib:
> Looks like latest alpha version is 4.2.0?
> 
> https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/044fd03c4a7d3b75aeb6ca2bd7fb6155d2cdb787
> 
> https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/044fd03c4a7d3b75aeb6ca2bd7fb6155d2cdb787%5E%21/#F0
> 
> not sure what I should target. latest and greatest?

I guess we should target the latest stable release available on Jetpack: https://developer.android.com/jetpack/androidx/releases/games (i.e. 4.0.0)

MarijnS95:
> fwiw, I intentionally didn't reply to that comment while it came across as kinda rude / accusatory. I have opened various upstream issues with GameActivity in the past.

I'm sorry, I thought the same and initially wanted to say something about it but that never materialized. Said something now, hope you didn't feel too alone/bothered when getting complaints for all your hard work.

---

> From what i was told, NativeActivity is deprecated and it was difficult to get the keyboard to work.

I don't think this can truly be deprecated because it's part of the core Android framework and always supposed to be there and "just work".

Yes, because it's hardlocked into the Android system it becomes impossible to forward-port bugfixes and new features, which is possible with `GameActivity`. But that thing is incredibly complex as we're showing here, not to mention getting it to integrate well with Rust code.

So far everyone (at least we) seem to be moving into the direction of making custom subclasses of `Activity` and/or `NativeActivity`, and have it abstract/interop _just the features we need_ all cleanly and natively implemented in Java/Kotlin or Rust, no external integration required (besides compiling and bundling a few `.class` files).

Unfortunately a common system for building and distributing that has not yet materialized, I'm working on some ideas (also being the main maintainer for `cargo-apk` and `xbuild`) but it's a bit hard to weed out all the NIH.

jb55:
> So far everyone (at least we) seem to be moving into the direction of making custom subclasses of Activity and/or NativeActivity, and have it abstract/interop just the features we need all cleanly and natively implemented in Java/Kotlin or Rust, no external integration required (besides compiling and bundling a few .class files).

are there any examples of this approach? or is this still just an idea

rib:
> So far everyone (at least we) seem to be moving into the direction of making custom subclasses of `Activity` and/or `NativeActivity`, and have it abstract/interop _just the features we need_ all cleanly and natively implemented in Java/Kotlin or Rust, no external integration required (besides compiling and bundling a few `.class` files).

yeah in an ideal world I still tend to think the best solution would be for us to create a `RustActivity` and then perhaps drop support for `GameActivity`. I was recently pondering this again briefly and considering taking a stab at this.

Although it addresses a number of issues in `NativeActivty`, `GameActivity` has been a bit of a pita in various ways and the larger amount of C/C++ code that is vendored (compared to the Rust port we did for `NativeActivity`) has been the source of multiple classic memory overflow / ownership bugs that should be more avoidable with a Rust implementation.

`NativeActivity` will probably always be a special case that should be supported because it's the only one that's part of the core Android SDK and is the only option for build systems that aren't able to pull in Java/Kotlin dependencies when packaging.

rib:
> are there any examples of this approach? or is this still just an idea

It's just a pipe dream still that we've discussed a bunch of times over the years

jb55:
had Claude do an analysis of how much work this would be. I just code2prompt'd the game-text-input codebase. Just leaving this here for my future self and others who are crazy enough to do this (but honestly doesn't seem like a bad idea)

https://claude.ai/share/5eb35755-917d-466c-a2ce-d8d9d06bdef0

MarijnS95:
My inner correctness exploded when reading how awfully wrong that AI output is. I need to recover from reading this.

rib:
> had Claude do an analysis of how much work this would be. I just code2prompt'd the game-text-input codebase. Just leaving this here for my future self and others who are crazy enough to do this (but honestly doesn't seem like a bad idea)
> 
> https://claude.ai/share/5eb35755-917d-466c-a2ce-d8d9d06bdef0

Btw, in the case of the game-text-input library specifically, that's probably one case where we wouldn't really want to just port the existing native side to Rust and we should instead aim for a different Java/Kotlin implementation too that lets us support an IME interface that's more expressive. One of the pain points with enabling IME in Winit on Android has specifically been with the limitations of the game-text-input IME interface.

the tedious part before that is just porting the C/C++ integration code to Rust but since I already did that for NativeActivity we could either use that as a starting point or at least a point of reference for a similar port for GameActivity.

jb55:
Should we target the Preedit/Commit stuff like in the winit IME apis? I haven't thought too much about that but I did find it weird that things did not map well over there.

rib:
> Should we target the Preedit/Commit stuff like in the winit IME apis? I haven't thought too much about that but I did find it weird that things did not map well over there.

exactly - the game-text-input api is a bit too simplistic, which makes it awkward to implement IME APIs on top of it in toolkits / Winit.

rib:
> > Should we target the Preedit/Commit stuff like in the winit IME apis? I haven't thought too much about that but I did find it weird that things did not map well over there.
> 
> exactly - the game-text-input api is a bit too simplistic, which makes it awkward to implement IME APIs on top of it in toolkits / Winit.

(though maybe the latest version is better - I haven't really checked recently)

MarijnS95:
> are there any examples of this approach? or is this still just an idea

Unfortunately an implementation of this is in a private repository. Not that it matters too much, I was put on a separate project right as someone else started implementing it in our codebase and despite my efforts to describe the ideal outcome that would be friendly to the entire Android community/crates/tooling we're building here at `rust-mobile`, none of that materialized.

It's not very complicated though. All I'll add is some very basic tooling in `build.rs` that compiles your Java/Kotlin and provides the resulting `.class` file (or check that in to the crate directly to cut down on dependencies) to one of the build tools to merge together (from all dependency crates) into a single dex file using `d8`, as described in the footnote here: https://github.com/MarijnS95/android-support

Secondly, that `RustActivity` will be a simple class like - or perhaps even subclassing - `NativeActivity`, and we can add more callbacks and interfaces whenever we wish while implementing the Java and Rust-JNI code ourselves (things to get access to the correct `SurfaceControl`, [receive intents and intent results](https://github.com/rust-mobile/android-intent), provide convenient helpers to [hide the UI](https://github.com/MarijnS95/android-support/blob/master/java/Activity.java), which is quite common for a game...). The possibilities are endless.

For end-users of `android-activity` this'll all be completely opaque, they'll just get access to more APIs directly from Rust without having to bother hacking around in JNI via some `Activity` pointer to do what they want because neither `NativeActivity`/`GameActivity`/NDK provide it. And everything that's contributed to this activity/crate will benefit the wider Rust Android system and probably be used directly from `winit`, which is now rather limited in what it can provide on Android.

This still needs a way to then subclass our `RustActivity` in case there's something _"really custom"_ people want to do in Java/Kotlin without having to copy half the setup.

Finally, this all neatly integrates with the new classes and Java/JNI interop that I'm adding to the `ndk` crate.

---

But yeah, as @rib already mentioned, we'd first have to bootstrap this whole activity before someone might want to start taking a stab at adding IME APIs to it (not me).

---

> the tedious part before that is just porting the C/C++ integration code to Rust but since I already did that for NativeActivity we could either use that as a starting point or at least a point of reference for a similar port for GameActivity.

@rib how "direct" was this C++-to-Rust port of `NativeActivity` glue? This also has its fair share of deadlocks (see open and merged PRs) and I'd rather use our own discretion to implement something that's sane. That should go hand in hand with fixing/improving the safe NDK wrapper primitives to make that possible though: https://github.com/rust-mobile/ndk/issues/478

MarijnS95:
https://github.com/rust-mobile/ndk/pull/498 @jb55 this is your team (or someone working on the same project at least), right? :joy:

jb55:
> https://github.com/rust-mobile/ndk/pull/498 @jb55 this is your team (or someone working on the same project at least), right? :joy: 

Lol yeah he's our pm he's a character

jb55:
Closing this in favor of:

- #191

PR #189, opened 2025-03-13
Update to gamesdk v4.0.0 with our rebased patches

jb55:
This is a WIP as I have not tested or compiled anything, I simply rebased our local patches (with proper formatting) on top of 4.0.0. It was surprisingly a clean rebase even though upstream moved things around and clang-formatted everything. I had to get somewhat clever with the patch rebasing and formatting to align things up.

I decided to replay all of the patches from gamesdk so that it can be verified that nothing was messed up and to not break bisectability. To do this I first reverted the original patches on 2.0.2, reformatted them, and then cleanly rebased them on top of 4.0.0.

I noticed some includes changed which might have broken stuff. I will start testing this next!


rib:
Cool - thanks for taking a pass at this!

It's good that this explicitly re-applies `GameActivity PATCH:` changes so it remains clear what integration changes were carrying.

I'm not so sure that we need to pull in the upstream history for GameActivity into the repo though by rebasing all of their patches. Previously we've just taken a vanilla snapshot of whatever version we want to support (replacing whatever we have in tree previously) and then applied our integration patches on that.

I guess we won't really gain the ability to bisect by pulling in their full history while we still require some patches on top that would also need to be rebased if you wanted to try bisecting through specific gamesdk commits?

If we use a merge commit with this PR that could help avoid having their history add noise to the `android-activity` history but am wondering if we should squash it.

Maybe you're guessing it might be useful to be able to rebase the integration patches to bisect any potential issues while testing?

jb55:
squashed!

PR #193, opened 2025-04-01
Bump rust-version to 1.73.0

rib:
bindgen-cli 0.71 (which we want to use) requires rustc >= 1.70

If we're bumping the rust version anyway it looks like it makes sense to at least bump to 1.73 which happens to fix the `stat` struct definition for Android.

Technically since we run bindgen-cli offline, their minimum version doesn't _need_ to affect us but I'd rather just sync and not track multiple "minimum" versions for this crate.

Rust 1.73 was released October 2023, which is still well over a year old and very conservative.

This updates `generate-bindings.sh` to pass `--rust-target 1.73.0` so we avoid generating bindings that require a more recent compiler.

### DIFF
--- a/android-activity/Cargo.toml
+++ b/android-activity/Cargo.toml
@@ -10,12 +10,10 @@ documentation = "https://docs.rs/android-activity"
 description = "Glue for building Rust applications on Android with NativeActivity or GameActivity"
 license = "MIT OR Apache-2.0"
 
-# 1.69 was when Rust last updated the Android NDK version used to build the
-# standard library which avoids needing the -lunwind workaround in build tools.
-#
-# We depend on cargo-ndk for building which has dropped support for the above
-# linker workaround.
-rust-version = "1.69.0"
+# Even though we could technically still build with 1.69, 1.73 has a fix for the
+# definition of the `stat` struct on Android, and so it seems worthwhile drawing
+# a line under that to ensure android-activity applications have that fix.
+rust-version = "1.73.0"
 
 [features]
 # Note: we don't enable any backend by default since features
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         # See top README for MSRV policy
-        rust-version: [1.69.0, stable]
+        rust-version: [1.73.0, stable]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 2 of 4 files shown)


rib:
Btw I'll probably merge eagerly, this without blocking on review, considering that this seems pretty uncontroversial and it should be fine to follow up with any issue if necessary in another PR (and I want to stack other changes / PRs on this).

android-activity/Cargo.toml
```diff
@@ -10,12 +10,9 @@ documentation = "https://docs.rs/android-activity"
...
 
-# 1.69 was when Rust last updated the Android NDK version used to build the
-# standard library which avoids needing the -lunwind workaround in build tools.
-#
-# We depend on cargo-ndk for building which has dropped support for the above
-# linker workaround.
-rust-version = "1.69.0"
+# We need a version > 1.70 for running bindgen-cli, and 1.73 also happens to
+# have a fix for the definition of the `stat` struct on Android.
+rust-version = "1.73.0"
```
MarijnS95, line 15: I don't think mentioning `bindgen-cli` is very relevant here (and I'd be more than happy to keep them separate). Having it in the PR description is more than enough.

MarijnS95 (approved):
Good one on setting `--rust-target` - doesn't that result in generated output changes though? Or was that only since `1.74`?

rib:
> Good one on setting `--rust-target` - doesn't that result in generated output changes though? Or was that only since `1.74`?

I guess rust-bindgen could change it's output for any major version, regardless of the `--rust-target` version that only needs to constrain that the generated code will be backwards-compatible with that compiler version.

I guess it's also valid to pass a `--rust-version` that's newer than whatever rustc version was used to build `bindgen-cli`.

I think it's OK to leave it to the PR that's updating to GameActivity 4.0.0 for re-generating new bindings. The bindings themeselves are private, so it's up to us to choose to re-generate if we want to take advantage of new bindgen features or when changing the native `GameActivity` interface.

rib:
> > but I'd rather just sync and not track multiple "minimum" versions for this crate.
> 
> I don't think we really need to "track" two versions; our developer-only build tools can just use whatever is the latest `stable`. At least I update my local Rust version as soon as it's released, and I think we're mostly the only two people contributing to this anyway.

yeah, fair enough - this version bump is technically not really needed.

any particular objection to me updating the rust-version to 1.73 anyway?

there is the `stat` struct fix for Android that is unlikely to affect many people but could still be a reason enough to set as a minimum version.

rib:
> No objections as long as it's not specifically on the merit of our developer tools/experience 👍

I'll update the commit message / comments etc to only refer to the `stat` struct fix and mostly forget about the interaction with rust-bindgen which is kinda irrelevant really while it's fine to use whatever latest toolchain we like when building/running bindgen-cli.

I'll keep the `--rust-target` change in `generate-bindings.sh` since that still makes sense in the context of a rust-version bump, and we probably should have passed this before (assuming the option existed before).

MarijnS95:
Yeah, love to see the `--rust-target` since that sets a proper boundary on what we can use.

PR #194, opened 2025-04-01
Re-export 'ndk' and 'ndk_sys' crates

rib:
Since we expose `ndk` types in the public API it makes sense to re-export these APIs so users of android-activity can defer to these without needing to manually sync the versions for explicit dependencies.

### DIFF
--- a/android-activity/src/lib.rs
+++ b/android-activity/src/lib.rs
@@ -119,12 +119,17 @@ use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::Duration;
 
-use input::KeyCharacterMap;
+use bitflags::bitflags;
 use libc::c_void;
+
 use ndk::asset::AssetManager;
 use ndk::native_window::NativeWindow;
 
-use bitflags::bitflags;
+// Since we expose `ndk` types in our public API it's convenient if crates can
+// defer to these re-exported APIs and avoid having to bump explicit
+// dependencies when they pull in new releases of android-activity.
+pub use ndk;
+pub use ndk_sys;
 
 #[cfg(not(target_os = "android"))]
 compile_error!("android-activity only supports compiling for Android");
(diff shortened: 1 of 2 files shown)


android-activity/src/lib.rs
```diff
@@ -124,6 +124,12 @@ use libc::c_void;
 use ndk::asset::AssetManager;
 use ndk::native_window::NativeWindow;
 
+// Since we expose `ndk` types in our public API it's convenient if crates can
+// defer to these re-exported APIs and avoid having to bump explicit
+// dependencies when they pull in new releases of android-activity.
+pub use ndk;
+pub use ndk_sys;
+
 use bitflags::bitflags;
```
MarijnS95, line 133: We should probably move this `use` to the "group" above.

MarijnS95 (approved):
Neat. I should really start looking into https://github.com/rust-mobile/ndk/issues/478 so that we can use and at the same time also return those internal `ndk` types back to users wherever needed/useful.

android-activity/src/lib.rs
```diff
@@ -124,6 +124,12 @@ use libc::c_void;
 use ndk::asset::AssetManager;
 use ndk::native_window::NativeWindow;
 
+// Since we expose `ndk` types in our public API it's convenient if crates can
+// defer to these re-exported APIs and avoid having to bump explicit
+// dependencies when they pull in new releases of android-activity.
```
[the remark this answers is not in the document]
rib: 'dependencies' is a surprisingly long word I suppose :)

I use a vscode plugin for wrapping at various column widths and this is already wrapped:
image

android-activity/src/lib.rs
rib: I moved this up and moved the `use input::KeyCharacterMap` down to the `mod input` definition.

PR #198, opened 2025-05-06
Expose main thread's looper for runOnUiThread compatibility

markkimsal:
Posting to the UI thread is only possible via the looper obtained from the main activity thread.  `android-activity` creates the `AndroidApp` inside a worker thread and there is no way to access the main thread's looper.  Even interacting with the UI will send events into the `AndroidAppInner` looper and execute code in the worker thread.

This PR simply records a pointer to the thread's looper before spawning the worker thread and makes it available in a similar fashion to the existing looper.

Related discussion:
https://github.com/rust-mobile/android-activity/discussions/197


rib:
Hi, sorry for taking a long time to look at this.

Thanks, the idea seems good to me.

I've rebased this and added corresponding support to the `game-activity` backend.

I've updated the implementation to expose `ndk::looper::ForeignLooper` instead of raw `ALooper` pointers to avoid any ambiguity with needing to retain/release `ALooper` references (and you can still use `.ptr()` if you want the underlying `ALooper` API.

The other change I made was to refer to this as the "java main" Looper and not just the "main looper" to avoid confusion with the thread/Looper that's associated with `android_main`.

Based on this work, I have also experimented with building a more ergonomic `AndroidApp::run_on_java_main_thread` API that can be used to run boxed closures on the java main / ui thread and I'm planning to open a follow PR for that soon.

PR #191, opened 2025-03-14
Update to GameActivity 4.0.0

jb55:
This updates the `game-activity` backend to be based on the latest `GameActivty` `4.0.0` release from Google's AGDK.

To make it easier to manage future updates to new versions of `GameActivity` the `android-activity` integration patches are now maintained under a fork of Google's games-sdk repo which can be found here:

https://github.com/rust-mobile/android-games-sdk/tree/android-activity-4.0.0

During development it's possible to build against an external games-sdk repo (as well we re-generate bindings) by setting the `ANDROID_GAMES_SDK` environment variable to point to a clone of the `android-games-sdk` repo.

To avoid having downstream developers need to deal with Git submodules, this adds an `import-games-sdk.sh` script that will copy the `prefab-src` code from the games-sdk into this repo.

- Closes https://github.com/rust-mobile/android-activity/pull/185
- Fixes https://github.com/rust-mobile/android-activity/issues/188
- Fixes https://github.com/rust-mobile/android-activity/issues/186

PR #200, opened 2025-06-30
GameActivity-4.0.0: add set_ime_editor_info

jb55:
This builds on top of:

- #191 

The commit in this PR is:

- https://github.com/jb55/android-activity/compare/set_ime_editor_info%5E...set_ime_editor_info

This adds support for `GameActivity_setImeEditorInfo` which is needed for setting various IME hints, such as multiline.

I'm currently using this in my egui/winit fork for mapping a new Multiline [ImePurpose](https://docs.rs/winit/latest/winit/window/enum.ImePurpose.html) to ime editor options:

- https://github.com/damus-io/winit/commit/eaff639ab0a14fccf595241f687be883154b267c

This ensures that the enter action button doesn't close the keyboard, it instead enters a newline.


nicoburns:
@jb55 What is the status of this PR? (I see it is still marked as a draft). I believe that the ability to set the `action_id` using this API (which isn't here, but looks like it'd be quite easy to add) + the ability to hook into the `onEditorAction` event (https://stackoverflow.com/questions/2004344/how-do-i-handle-imeoptions-done-button-click) are what I need in order to properly handle the "enter key" on android soft keyboard.

rib:
For reference `set_ime_editor_info()` now also accepts an `action: TextInputAction` so the behaviour of the enter/action key is configurable (addressing a remaining `TODO` comment in the code)

